PR #1, opened 2022-07-15
Consistency improvements

lumburovskalina:


src/app/weak-supervision/components/weak-supervision/weak-supervision.component.ts
```diff
@@ -490,4 +492,11 @@ export class WeakSupervisionComponent implements OnInit, OnDestroy {
     })
 
   }
+
+  navigateToSettings() {
+    this.router.navigate(["../settings"], {
+      relativeTo: this.activatedRoute,
+    });
+    localStorage.setItem("openModal", "true");
```
JWittmeyer: i think i would switch these so the localstorage is set before the route is changed to prevent some race condition

JWittmeyer (approved):

JWittmeyer:
on heuristics when clickling on one of the labeling task filter (or next to them) an error is thrown
image

- [x] resolved

JWittmeyer:
on lookuplists clicking to create a new list shows an error. Only if no lists exist yet

image


- [x] resolved

PR #10, opened 2022-07-25
Shared component dropdown

lumburovskalina:
Shared components for dropdown with fixes values and iterative dropdown (also with checkboxes).
Replaced code with the shared components.


src/app/base/components/dropdown-iterative/dropdown-iterative.component.ts
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+
+  getDropdownDisplayText(
+    formControls: AbstractControl[],
+    labelFor: string
+  ): string {
+    let text = '';
+    let atLeastOneNegated: boolean = false;
+    for (let c of formControls) {
```
JWittmeyer: does this only work with a list of form controls? not with a list of strings?

src/app/base/components/dropdown-iterative/dropdown-iterative.component.ts
```diff
@@ -0,0 +1,93 @@
...
+})
+export class DropdownIterativeComponent implements OnInit {
+
+  @Input() arrayOptions = [];
+  @Input() condition;
+  @Input() valueIfConditionTrue;
+  @Input() valueIfConditionFalse;
+  @Input() buttonTooltip: string;
+  @Input() labelingTasks;
+  @Input() property: string;
```
JWittmeyer: not really descriptive. i wouldn't know what property this is or what this is for

src/app/base/components/dropdown-iterative/dropdown-iterative.component.ts
```diff
@@ -0,0 +1,93 @@
...
+  @Input() optionProperty: string;
+  @Input() disabledCondition: boolean;
+  @Input() hasCheckboxes: boolean;
+  @Input() formArrayName: string;
```
JWittmeyer: i think we can reduce the amount of input parameter by using a type with optional parameters that hold the specification. I think something like " @Input() dropdownValues = []; @Input() dropdownOptions;" options in turn have values for "hasCheckbox?:boolean, useFormArray?:FormArray, svgArrag?:string[]...) so the basic version would only use the list of arrays but additional logic can be added via the array.

If it helps I could build a lite version to better show what I mean

src/app/base/components/dropdown/dropdown.component.ts
```diff
@@ -0,0 +1,181 @@
+import { Component, EventEmitter, Input, OnDestroy, OnInit, Output } from '@angular/core';
```
JWittmeyer: as mentioned before, this feels very hard to maintain being completely hardcoded. i think to reduce our stress in the long run we should aim for a more flexible solution (like with the dynamic dropdown)

JWittmeyer:
Distances for sample projects dont match (guessing its because of the missing picture element for the initial cases)
image

- [x] resolved

JWittmeyer:
run selected (on heuristics) isn't disabled when no heuristic is selected

- [x] resolved

JWittmeyer:
dropdown of lookuplist actions is bigger than the screen.
image

- [x] resolved

JWittmeyer:
Size of checkbox dropdown seems to have changed:
dev.app:
image

local:
image

- [x] resolved

src/app/base/components/dropdown-iterative/dropdown-iterative.component.ts
lumburovskalina: I will remove both dropdown and dropdown iterative components and then replace dropdown-it with dropdown.

src/app/base/components/dropdown/dropdown-helper.ts
```diff
@@ -0,0 +1,38 @@
...
+ * @buttonTooltip {string, optional} - adds a tooltip if defined
+ * @isDisabled {boolean, optional} - disables the dropdown
+ * @isOptionDisabled {boolean[], optional} - disables the dropdown option (needs to be the exact same length as the optionArray)
+ * @optionIcons {string[], optional} - displays a predfined icon if set for the index (needs to be the exact same length as the optionArray)
+ * @hasCheckboxes {boolean, optional} - helper for checkbox like dropdowns (e.g. data browser)
+ * @buttonVersion {string, optional} - defaults to 'default' (button with a caption text), '...', 'userIcon'
+ * @avatarUri {string, optional} - link to the avatar image for logged user
+ * @prefix {string, optional} - prefix to the main name in the option
+ * @postfix {string, optional} - postfix to the main name in the option
+ * @buttonHasBlueStyle {boolean, optional} - flag if the button has the blue style
```
JWittmeyer: this isn't really clear to me, what is the "blue" style?

lumburovskalina: The button for 'Sample projects'

src/app/knowledge-bases/knowledge-base-details/knowledge-base-details/knowledge-base-details.component.ts
```diff
@@ -328,4 +316,20 @@ export class KnowledgeBaseDetailsComponent implements OnInit, AfterViewInit, OnD
...
+      case 'Edit term': 
+        this.openTermEditor(true, term.id, term.value, term.comment);
+        break;
+      case 'Remove term':
+        this.deleteTerm(term.id);
+        break;
+      case 'Blacklist term':
+      case 'Whitelist term':
+        this.blackListTerm(term.id);
+        break;
```
JWittmeyer: from the function names and the case this looks like a mistake. Maybe the function `this.blackListterm`... should be changed to `this.toggleBlacklistTerm` to reduce confusion. (I know that the backend also works with a toggle even though the name isn't representative of it but we need to start somewhere :) )

src/app/base/components/dropdown/dropdown-helper.ts
JWittmeyer: yes but is it a blue tint or is it a "special" option like hasCheckboxes that has custom logic only meant for that specific usecase?

JWittmeyer: also doesnt seem to work as this is a different color than on the dev app:
current:
image

dev app:
image


also I think the size is different (test size)

lumburovskalina: No, it is for all buttons that we want to have this style.

src/app/knowledge-bases/knowledge-base-details/knowledge-base-details/knowledge-base-details.component.ts
lumburovskalina: Yes sure, changed

src/app/base/components/dropdown/dropdown-helper.ts
lumburovskalina: Yes, I saw that but I thought all drodowns should have the same style, if this one is so different that the others does it make sense to be replaced with the shared component at all? Or we can add condition to use specific characteristics for this dropdown.

JWittmeyer: i think i'd go with options for 
1. hovertint or hovercolor (gray = default, blue = this blue & hexcode for everything else)
2. text color (default = black & hexcode for everything else)
3. text hover color (default = white & hexcode for everything else)
4. text size (default = text-sm (class names possible))

JWittmeyer:
i thought we fixed the width thing however it doesn't seem like we did since now we have it as exact width,
image

I tested a bit around and the dropdown menu with min width 100% and the element with width:max-content should do it (hopefully :D )

- [x] resolved

JWittmeyer (approved):

src/app/data/components/data-browser/data-browser.component.html
```diff
@@ -842,9 +786,9 @@
                         Clear filter
                     </button>
                 </div>
-                
 
-                <div data-tip="Change the data browser configurations" class="tooltip tooltip-left relative flex items-center"
+                <<<<<<< HEAD <div data-tip="Change the data browser configurations"
```
JWittmeyer: i think i f... something up with the merge :D <<<<<<< HEAD shouldn't be there

lumburovskalina: Removed it.

PR #11, opened 2022-07-26
Discord link and icon on sidebar menu

### LINKED ISSUE #8: Add Discord link on sidebar
Add an icon and a suitable link to Discord on the sidebar menu.

lumburovskalina:
Added discord icon and link on sidebar.
Fix in data browser with z-index.

### DIFF
--- a/src/app/base/components/sidebar-pm/sidebar-pm.component.html
+++ b/src/app/base/components/sidebar-pm/sidebar-pm.component.html
@@ -34,7 +34,7 @@
 
                     <div *ngIf="project$ | async as project">
                         <div *ngIf="project.id !== null">
-                            <div class="flex items-center justify-center overflow-visible mt-12">
+                            <div class="flex items-center justify-center overflow-visible mt-10 2xl:mt-12">
                                 <div data-tip="Overview" class="tooltip tooltip-right relative z-50"
                                     [ngClass]="project.numDataScaleUploaded == 0 ? 'opacity-50 cursor-not-allowed pointer-events-none' : 'opacity-100 cursor-pointer'">
                                     <a [routerLink]="['./overview']" rel="noopener noreferrer" class="circle text-white"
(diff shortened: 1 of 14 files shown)


src/app/data/components/data-browser/data-browser.component.scss
```diff
@@ -87,5 +87,5 @@ input[type="number"] {
 }
 
 .tooltip-label::before {
-  z-index: 10;
+  z-index: 50;
```
JWittmeyer, line 90: if we want to use our own class for this why not also include the display relative part which is a prerequisite for this to work?

src/app/base/components/sidebar-pm/sidebar-pm.component.html
```diff
@@ -153,6 +153,23 @@
                             </a>
                         </div>
                     </div>
+                    <div class="flex items-center justify-center overflow-visible mt-12">
+                        <div data-tip="Join our community" class="tooltip tooltip-right relative z-50"
+                            data-tooltip-target="tooltip-right" data-tooltip-placement="right">
```
JWittmeyer: we sometimes use data-tooltip-target and sometimes class tooltip-right. As far as I know, only the tooltip-right is needed to place it right. what is the other part for?

JWittmeyer:
we are putting more and more icons in the sidebar. Should we maybe include a scroll option if the screen is too small? That way it wouldn't be super pretty but at least accessible. Current min height around 700 px screen size.
image

- [x] resolved
- [ ] ignored

src/app/base/components/sidebar-pm/sidebar-pm.component.html
lumburovskalina: I think the data-tooltip-target should be removed on all places, I removed the part and it nothing is changed since the tooltip-right does it work. If it okay I will remove it everywhere as part of the PR.

src/app/data/components/data-browser/data-browser.component.scss
lumburovskalina: I think there is no need for own class tooltip-label, we can use the tooltip and add the z-index. there. I will replace it as part of this PR.

src/app/base/components/sidebar-pm/sidebar-pm.component.html
JWittmeyer: lets go for it

JWittmeyer (approved):

PR #12, opened 2022-07-26
Saving forever fix and opacity tooltip fixed

### LINKED ISSUE #7: LF or AL changes seem to save forever
refreshing the site shows that the data was saved to the backend but the saving animation (and run disable) stays until a manual refresh was used

lumburovskalina:
Fixed saving on labeling function and active learning pages from loading forever.
Opacity on button Run + weakly supervised fixed.

### DIFF
--- a/src/app/weak-supervision/components/weak-source-details/weak-source-details.component.ts
+++ b/src/app/weak-supervision/components/weak-source-details/weak-source-details.component.ts
@@ -424,7 +424,7 @@ export class WeakSourceDetailsComponent
           this.description,
           functionName
         ).pipe(first())
-        .subscribe();
+        .subscribe(() => this.informationSourceQuery$.refetch());
     } else {
       console.log('currently not possible for type ' + sourceType);
       return;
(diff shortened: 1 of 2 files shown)


src/app/weak-supervision/components/weak-source-details/weak-source-details.component.ts
```diff
@@ -424,7 +424,7 @@ export class WeakSourceDetailsComponent
           this.description,
           functionName
         ).pipe(first())
-        .subscribe();
+        .subscribe(() => this.informationSourceQuery$.refetch());
```
JWittmeyer: shouldn't be necessary because of the websockets (or at least that's the theory, ill double check)

JWittmeyer: it was only updated for different user ids (don't know why though).

Also refetching by hand on mutations isn't a common approach, usually, we would use refetch queries in apollo (however it's already there so no clue why it didn't work).

Can be resolved.

JWittmeyer (approved):

PR #13, opened 2022-07-26
Long embedding names that break modal for zero-shot fixed

### LINKED ISSUE #6: Long embeddings names (Huggingface models) can break the modal on heuristics
Bildschirmfoto 2022-07-21 um 16 22 05

div with modal-box should include something like
```css
max-width: 80%; 
min-width: 55%;
width:fit-content;
```

lumburovskalina:
Fixed long embedding names that break the modal for zero-shot

### DIFF
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -114,6 +114,10 @@ module.exports = {
       "xxs": "9rem",
       "xs-1": "250px",
       ...defaultTheme.maxWidth,
+      "3/5": "80%"
+    },
+    minWidth: {
+      "1/2": "50%"
     },
     lineHeight: {
       '12': '3rem',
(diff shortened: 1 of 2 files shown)


tailwind.config.js
```diff
@@ -114,6 +114,10 @@ module.exports = {
       "xxs": "9rem",
       "xs-1": "250px",
       ...defaultTheme.maxWidth,
+      "3/5": "80%"
+    },
+    minWidth: {
+      "1/2": "50%"
```
JWittmeyer: why was this added to the tailwind config? and not e.g. a scss class for the needed logic?

lumburovskalina: We can use this property for all modals, so I assumed it will be better if we have it as a 'class' instead of writing it as a style.

JWittmeyer (approved):

PR #15, opened 2022-07-26
Heuristic statuses shared component

### LINKED ISSUE #14: Create a shared component for statuses
Shared component for statuses (Initial, Created, Finished, Error)
Currently duplicate code is used on different pages and with the shared component that can be changed.

Changes impact the following pages: heuristics page, heuristic details page and zero-shot page.

lumburovskalina:
Shared component for statuses (Initial, Created, Finished, Error)
Currently duplicate code is used on different pages and with the shared component that can be changed.
Changes impact the following pages: heuristics page, heuristic details page and zero-shot page.
Tooltips fixed on zero-shot(when the buttons are disabled the tooltips should still appear).
Renamed items to records on heuristics page.

### DIFF
--- a/src/app/base/components/heuristic-statuses/heuristic-statuses.component.html
+++ b/src/app/base/components/heuristic-statuses/heuristic-statuses.component.html
@@ -1,0 +1,28 @@
+<ng-container *ngIf="status !== 'CREATED' && status !== 'FINISHED' && status !== 'FAILED'"
+    [ngTemplateOutlet]="heuristicStatus" 
+    [ngTemplateOutletContext]="{dataTip:'Heuristic was successfully registered.', color: 'gray',statusName:'Initial'}"></ng-container>
+
+<ng-container *ngIf="status === 'CREATED'"
+    [ngTemplateOutlet]="heuristicStatus" 
+    [ngTemplateOutletContext]="{dataTip:'Heuristic is currently being executed.', color: 'yellow',statusName:'Running'}"></ng-container>
+
+<ng-container *ngIf="status === 'FINISHED'"
+    [ngTemplateOutlet]="heuristicStatus" 
+    [ngTemplateOutletContext]="{dataTip:'Heuristic was successfully executed.', color: 'green',statusName:'Finished'}"></ng-container>
+
+<ng-container *ngIf="status === 'FAILED'"
+    [ngTemplateOutlet]="heuristicStatus" 
+    [ngTemplateOutletContext]="{dataTip:'Heuristic ran into errors.', color: 'red',statusName:'Error'}"></ng-container>
+
+
+<ng-template #heuristicStatus let-dataTip="dataTip" let-color="color" let-statusName="statusName">
+    <div [attr.data-tip]="dataTip" class="tooltip ml-2" [ngClass]="tooltipPosition">
+        <span
+            class="inline-flex items-center px-2 py-0.5 rounded text-xs font-medium" [ngClass]="[color === 'green' ? 'bg-green-300' :'bg-'+color+'-100','text-'+color+'-800']">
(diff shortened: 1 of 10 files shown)


src/app/base/components/heuristic-statuses/heuristic-statuses.component.html
```diff
@@ -0,0 +1,28 @@
+<ng-container *ngIf="status !== 'CREATED' && status !== 'FINISHED' && status !== 'FAILED'"
```
JWittmeyer: out of curiosity why are these with type check (=== vs ==)?

src/app/base/components/heuristic-statuses/heuristic-statuses.component.html
```diff
@@ -0,0 +1,28 @@
...
+    [ngTemplateOutlet]="heuristicStatus" 
+    [ngTemplateOutletContext]="{dataTip:'Heuristic is currently being executed.', color: 'yellow',statusName:'Running'}"></ng-container>
+
+<ng-container *ngIf="status === 'FINISHED'"
+    [ngTemplateOutlet]="heuristicStatus" 
+    [ngTemplateOutletContext]="{dataTip:'Heuristic was successfully executed.', color: 'green',statusName:'Finished'}"></ng-container>
+
+<ng-container *ngIf="status === 'FAILED'"
+    [ngTemplateOutlet]="heuristicStatus" 
+    [ngTemplateOutletContext]="{dataTip:'Heuristic ran into errors.', color: 'red',statusName:'Error'}"></ng-container>
```
JWittmeyer: is there a specific reason why we solve this with template outlets instead of assigning the necessary variables in the oninit or onchange functions of the component?

Would reduce the size of the HTML quite a lot & might make it easier to read because of proper highlighting in ts file. Also switch could be utilized there without any issues.

src/app/base/components/heuristic-statuses/heuristic-statuses.component.ts
```diff
@@ -0,0 +1,18 @@
+import { Component, Input, OnInit } from '@angular/core';
+
+@Component({
+  selector: 'kern-heuristic-statuses',
+  templateUrl: './heuristic-statuses.component.html',
+  styleUrls: ['./heuristic-statuses.component.scss']
+})
+export class HeuristicStatusesComponent implements OnInit {
```
JWittmeyer: if its not used why is OnInit implemented?

lumburovskalina: Removed

src/app/base/components/heuristic-statuses/heuristic-statuses.component.html
lumburovskalina: No specific reason for here, to check the type and the value but in most cases we use the == operator, so maybe it would be better if we change it here so we can maintain consistency.

src/app/base/components/heuristic-statuses/heuristic-statuses.component.html
lumburovskalina: Moved the login in a switch case in the .ts file

src/app/base/components/heuristic-statuses/heuristic-statuses.component.ts
```diff
@@ -0,0 +1,43 @@
...
+  @Input() status: string;
+  @Input() tooltipPosition: string = 'tooltip-bottom';
+
+  dataTip: string;
+  statusName: string;
+  color: string;
+
+  constructor() { }
+
+  ngOnInit(): void {
```
JWittmeyer: i think this is something that can change on runtime so onChanges might be a better idea

JWittmeyer (approved):

PR #36, opened 2022-08-16
Fixes Bugs in the record ide

FelixKirsch:
Solves two issues:
- code is not executed when navigating to the next or previous record,
- if "import easteregg" is stored in local storage, snake is executed every time and one cannot return to the record ide.

### DIFF
--- a/src/app/record-ide/components/record-ide.component.ts
+++ b/src/app/record-ide/components/record-ide.component.ts
@@ -42,7 +42,6 @@ export class RecordIDEComponent implements OnInit {
     currentPos: number
   };
   snakeActive: boolean = false;
-  firstVisit: boolean = true;
   vertical: boolean = true;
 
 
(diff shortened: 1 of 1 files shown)


src/app/record-ide/components/record-ide.component.ts
```diff
@@ -102,16 +101,13 @@ export class RecordIDEComponent implements OnInit {
...
-        this.firstVisit = false;
-      }
+      this.runRecordIde(true);
     }));
     return this.project$.pipe(first());
   }
 
-  runRecordIde() {
-    if (this.code.indexOf("import easteregg") != -1) {
+  runRecordIde(firstVisit: bool = false) {
```
lumburovskalina: Is there a specific reason why we use bool and not boolean? I think we use boolean through the application.

lumburovskalina (approved):

src/app/record-ide/components/record-ide.component.ts
FelixKirsch: no, i will change it

PR #44, opened 2022-08-29
Model download page

lumburovskalina:
works with:
- https://github.com/code-kern-ai/refinery-embedder/pull/9
- https://github.com/code-kern-ai/refinery-config/pull/6
- https://github.com/code-kern-ai/refinery-zero-shot/pull/6
- https://github.com/code-kern-ai/refinery-gateway/pull/32
- https://github.com/code-kern-ai/dev-setup/pull/15


src/app/model-download/components/model-download-component/model-download-component.component.html
```diff
@@ -0,0 +1,231 @@
...
+                            <td class="text-center px-3 py-2 text-sm text-gray-500">
+                                <a [href]="model.link" target="_blank" *ngIf="model.link" class="h4 w-4 m-auto block"
+                                    style="padding:0px"><svg xmlns="http://www.w3.org/2000/svg" class="h-4 w-4 m-auto"
+                                    fill="none" viewBox="0 0 24 24" stroke="currentColor" stroke-width="2">
+                                    <path stroke-linecap="round" stroke-linejoin="round"
+                                        d="M10 6H6a2 2 0 00-2 2v10a2 2 0 002 2h10a2 2 0 002-2v-4M14 4h6m0 0v6m0-6L10 14" />
+                                </svg></a>
+                            </td>
+                            <td class="whitespace-nowrap text-center px-3 py-2 text-sm text-gray-500">
+                                <ng-template [ngIf]="model.date != 0 && (model.status === 'finished' || model.status === 'downloading')">{{parseUTC(model.date)}}</ng-template>
```
JWittmeyer: functions in html code are run quite often. changing the underlying donwloadedModels variable to be fully prepared on subscription would reduce workload (new field `parsedDate` or something)

src/app/model-download/components/model-download-component/model-download-component.component.html
```diff
@@ -0,0 +1,231 @@
...
+                            <th scope="col"
+                                class="px-3 py-2 text-center text-xs font-medium uppercase tracking-wide text-gray-500">
+                                Status</th>
+                            <th scope="col"
+                                class="px-3 py-2 text-center text-xs font-medium uppercase tracking-wide text-gray-500">
+                            </th>
+                        </tr>
+                    </thead>
+                    <tbody class="divide-y divide-gray-200">
+                        <tr *ngFor="let model of downloadedModels; let i = index">
```
JWittmeyer: is the variable i needed?

src/app/model-download/components/model-download-component/model-download-component.component.html
```diff
@@ -0,0 +1,231 @@
...
+                                        <line x1="18" y1="12" x2="21" y2="12"></line>
+                                        <line x1="16.25" y1="16.25" x2="18.4" y2="18.4"></line>
+                                        <line x1="12" y1="18" x2="12" y2="21"></line>
+                                        <line x1="7.75" y1="16.25" x2="5.6" y2="18.4"></line>
+                                        <line x1="6" y1="12" x2="3" y2="12"></line>
+                                        <line x1="7.75" y1="7.75" x2="5.6" y2="5.6"></line>
+                                    </svg>
+                                </div>
+                            </td>
+                            <td class="whitespace-nowrap text-center px-3 py-2 text-sm text-gray-500">
```
JWittmeyer, line 110: creating a delete modal for every row works but stresses the dom. For now this is fine, in the future we should try to use one modal at the end of the component & a variable reference for better readability and less dom clutter (see project settings #labelEditCheckbox and the related modal for reference)

src/app/model-download/components/model-download-component/model-download-component.component.html
```diff
@@ -0,0 +1,231 @@
...
+                                                (click)="selectEmbeddingHandle(modelHandle,inputEmbeddingHandle)">
+                                                {{modelHandle.configString}}
+                                            </label>
+                                            <!-- <svg xmlns="http://www.w3.org/2000/svg" style="margin: auto 0 auto auto" *ngIf=""
+                                                class="icon icon-tabler icon-tabler-file-check" width="24" height="24" viewBox="0 0 24 24" stroke-width="2" stroke="currentColor" fill="none" stroke-linecap="round" stroke-linejoin="round">
+                                                <path stroke="none" d="M0 0h24v24H0z" fill="none"></path>
+                                                <path d="M14 3v4a1 1 0 0 0 1 1h4"></path>
+                                                <path d="M17 21h-10a2 2 0 0 1 -2 -2v-14a2 2 0 0 1 2 -2h7l5 5v11a2 2 0 0 1 -2 2z"></path>
+                                                <path d="M9 15l2 2l4 -4"></path>
+                                             </svg> -->
```
JWittmeyer: comment in code

src/app/model-download/components/model-download-component/model-download-component.component.ts
```diff
@@ -0,0 +1,172 @@
...
+    [this.downloadedModelsQuery$, this.downloadedModelsList$] = this.informationSourceApolloService.getModelProviderInfo();
+    this.prepareModels();
+    this.prepareZeroShotRecommendations(this.projectId);
+    this.isManaged = ConfigManager.getIsManaged();
+
+    this.subscriptions$.push(
+      this.downloadedModelsList$.subscribe((downloadedModels) => this.downloadedModels = downloadedModels));
+
+    NotificationService.subscribeToNotification(this, {
+      projectId: this.projectId,
```
JWittmeyer: project-specific subscription might run into issues (see gateway info)

src/app/model-download/components/model-download-component/model-download-component.component.ts
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  initForm() {
+    this.form = this.formBuilder.group({
+      name: ['', Validators.required]
+    })
+  }
+
+  deleteModel(name: string, revision: string) {
+    this.subscriptions$.push(
```
JWittmeyer: no need to push this to subscriptions since the pipe(first()) unsubscribes on it's own

src/app/model-download/components/model-download-component/model-download-component.component.ts
```diff
@@ -0,0 +1,172 @@
...
+      this.informationSourceApolloService
+      .deleteModel(name,revision)
+      .pipe(first()).subscribe()
+    );
+  }
+
+  downloadModel() {
+    if(this.form.invalid) return ;
+    this.subscriptions$.push(
```
JWittmeyer: see above

src/app/model-download/components/model-download-component/model-download-component.component.ts
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  parseUTC(utc: any) {
+    const milliseconds = +utc * 1000;
+    const utcDate = dateAsUTCDate(new Date(milliseconds));
+    return utcDate.toLocaleString();
+  }
+
+  prepareModels() {
+    this.subscriptions$.push(this.projectApolloService.getRecomendedEncodersForEmbeddings(this.projectId)
```
JWittmeyer: not sure why this is used without pipe(first()) since the recommendations are hardcoded and shouldn't change over time

src/app/model-download/components/model-download-component/model-download-component.component.ts
```diff
@@ -0,0 +1,172 @@
...
+      .subscribe((models) => {
+        this.models = models.filter(el =>
+          el.configString != 'bag-of-characters' && el.configString != 'bag-of-words' && el.configString != 'tf-idf');
+      }));
+  }
+
+  prepareZeroShotRecommendations(projectId: string) {
+    let q, vc;
+    [q, vc] = this.informationSourceApolloService.getZeroShotRecommendations(projectId);
+    this.subscriptions$.push(vc.pipe(first()).subscribe((r) => {
```
JWittmeyer: subscription push with pipe first

src/app/model-download/components/model-download-component/model-download-component.component.ts
```diff
@@ -0,0 +1,172 @@
...
+      r.forEach((rec) => this.models.push(rec))
```
JWittmeyer: i think this might run into issues since this pushed to models while the other prepare function filters and replaces -> depending on the order of the return the result isn't deterministic

src/app/model-download/components/model-download-component/model-download-component.component.ts
```diff
@@ -0,0 +1,172 @@
...
+    }));
+  }
+
+  formatBytes(bytes, decimals = 2) {
+    if (bytes === 0) return '0 Bytes';
+    const dm = decimals < 0 ? 0 : decimals;
+    const sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
+    const i = Math.floor(Math.log(bytes) / Math.log(1024));
+    return parseFloat((bytes / Math.pow(1024, i)).toFixed(dm)) + ' ' + sizes[i];
+  }
```
JWittmeyer: can be put into helper-functions file (like dateAsUTCDate) and imported & used so changes only need to be done at one place

src/app/model-download/components/model-download/model-download.component.html
```diff
@@ -0,0 +1 @@
+<kern-model-download-component></kern-model-download-component>
```
JWittmeyer: i think the file path might be misleading, this is placed in model-download/components but it's the page, right?

src/app/projects/components/project-settings/project-settings.component.html
```diff
@@ -471,7 +484,7 @@
                                             (mouseenter)="setCurrentEmbeddingHandle(embeddingHandle,hoverBox,me)"
                                             (mouseleave)="setCurrentEmbeddingHandle(null)">
 
-                                            <label class="cursor-pointer label"
+                                            <label class="cursor-pointer label" [ngClass]="checkIfModelIsDownloaded(embeddingHandle.configString) ? 'text-green-700' : ''"
```
JWittmeyer, line 489: function in html code

src/app/projects/components/project-settings/project-settings.component.ts
```diff
@@ -117,6 +118,9 @@ export class ProjectSettingsComponent implements OnInit, OnDestroy, AfterViewIni
   }
   labelMap: Map<string, []> = new Map<string, []>();
   @ViewChild('modalInput', { read: ElementRef }) myModalnewRecordTask: ElementRef;
+  downloadedModelsList$: any;
+  downloadedModelsQuery$: any;
+  downloadedModels: any[] = [];
```
JWittmeyer: if this is set on every subscription this doesn't need to be initialized with []

src/app/projects/components/project-settings/project-settings.component.html
JWittmeyer: (works for now so we dont change it but we need to keep an eye on this in the future)

src/app/model-download/components/model-download-component/model-download-component.component.ts
lumburovskalina: I added the functions one after another to be executed, would that be a possible solution?

src/app/projects/components/project-settings/project-settings.component.html
lumburovskalina: Added this as an item to the refactoring ideas

src/app/model-download/components/model-download/model-download.component.html
lumburovskalina: Added it in a folder pages and the component stays in a folder components

src/app/model-download/components/model-download-component/model-download-component.component.html
lumburovskalina: I added it at the end of the component and set the value on click of the delete icon

src/app/model-download/components/model-download-component/model-download-component.component.ts
JWittmeyer: not sure since i cant see it yet. If we want to request simultaneously we could store the result of each in a variable and on either subscription return build a "combined" list. Requesting one after another should work however it might result in slightly longer wait (not noticeable with the amount of data I think)

JWittmeyer:
image

For the os version this page isn't really usable. Let's disable the link and add a data-tip for the os users "check out our hosted version to use this function"

- [x] resolved

JWittmeyer:
image

Link seems our of alignment
Same seems to be the case for delete or inside the delete button

- [x] resolved

src/app/model-download/components/model-download-component/model-download-component.component.ts
lumburovskalina: changed with a combined list

JWittmeyer:
image

Info seems to be missing for zero shot models.
Also i think it might be a good idea to differentiate between zero shot and embedding recommendations (e.g. by a divider line)

- [x] resolved

JWittmeyer:
Zero-shot recommendations aren't displayed ordered by prio

- [x] resolved

JWittmeyer:
zero shot test area "downloaded models" doesn't seem to be vertically aligned, same seems to be true for settings page

image

- [x] resolved

JWittmeyer:
hover on zero shot test area dropdown is still the same green
image

I think a lighter shade of green would be good here. Maybe lime?

- [x] resolved

JWittmeyer:
Sometimes on the zero shot page i get this error:
image

Not sure when it appears. I currently have 1 zero shot and one embedding model downloaded.
When it does appear the highlighting (green color) in the dropdown is missing

might be related to a race condition (one query is faster than expected)

- [x] resolved

src/app/base/components/dropdown/dropdown.component.html
```diff
@@ -37,7 +37,7 @@
         role="menu" aria-orientation="vertical" aria-labelledby="menu-button" tabindex="-1">
         <div class="py-1 cursor-pointer" role="none">
             <div *ngFor="let caption of dropdownOptionCaptions;let i = index" (click)="performActionOnOption($event,i)"
-                [ngClass]="[dropdownOptions.textSize,dropdownOptions.textColor,dropdownOptions.optionDescriptions ? '' : 'flex' ,dropdownOptions.isOptionDisabled?.length && dropdownOptions.isOptionDisabled[i]?'opacity-50 cursor-not-allowed':'',!dropdownOptions.hasCheckboxes ? dropdownOptions.hoverColor + ' ' +dropdownOptions.textHoverColor:'']"
+                [ngClass]="[dropdownOptions.textSize,dropdownOptions.textColor,dropdownOptions.optionDescriptions ? '' : 'flex' ,dropdownOptions.isOptionDisabled?.length && dropdownOptions.isOptionDisabled[i]?'opacity-50 cursor-not-allowed':'',!dropdownOptions.hasCheckboxes ? dropdownOptions.hoverColor + ' ' +dropdownOptions.textHoverColor:'', dropdownOptions.isModelDownloaded?.length && dropdownOptions.isModelDownloaded[i] ? 'text-green-700 hover:text-green-700': '']"
```
JWittmeyer: why do we use a new field for this and not the already implemented text color options?

src/app/projects/components/project-settings/project-settings.component.ts
```diff
@@ -173,6 +184,7 @@ export class ProjectSettingsComponent implements OnInit, OnDestroy, AfterViewIni
         }
       })
     }
+    this.isManaged = ConfigManager.getIsManaged();
```
JWittmeyer: i think we need to wait for the configmanager to be initialized first ngOnInit is usually to early

JWittmeyer:
the opacity of the data tip looks a bit weird (it's using the one from the button)
for an easy fix let's remove the opacity from the button for disabled (the not allowed cursor is enough here)

image

- [x] resolved

src/app/zero-shot-details/component/zero-shot-details.component.ts
```diff
@@ -110,6 +116,15 @@ export class ZeroShotDetailsComponent
...
     combineLatest(tasks$).subscribe(() => this.prepareInformationSource(projectId));
 
+    [this.downloadedModelsQuery$, this.downloadedModelsList$] = this.informationSourceApolloService.getModelProviderInfo();
+    this.subscriptions$.push(
+      this.downloadedModelsList$.subscribe((downloadedModels) => {
+        this.downloadedModels = downloadedModels;
+        this.createModelsDownloadedStateList();
+      }));
+
+    this.isManaged = ConfigManager.getIsManaged();
```
JWittmeyer: see above (config manager is requested before the init is done)

JWittmeyer: to prevent this issue in the future I added an error throw to the config manager if the value is requested before it's initialized ([commit link](https://github.com/code-kern-ai/refinery-ui/pull/44/commits/cb1d7678c49fdbb64c37df4d205210c22fa5105a))

src/app/base/components/dropdown/dropdown.component.html
JWittmeyer: just realized that the text color is a single value and not an array :D 

Thats the reason --> can be resolved

src/app/zero-shot-details/component/zero-shot-details.component.ts
JWittmeyer: error thow still has a bug, fixing it atm

JWittmeyer: check is now included (hopfully :D ) - its a console log now

lumburovskalina: Also the check if init is added so I am resolving the conversation :)

JWittmeyer (approved):

PR #57, opened 2022-09-18
clicking on weak supervision results adds them to manual label set

jhoetter:
Clicking on a weakly supervised label now adds them to the manual labeling set. Addresses code-kern-ai/refinery#51.

Next step could possibly be to tab-navigate through weak supervision results, and accepting them via enter.

### DIFF
--- a/src/app/labeling/components/labeling.component.ts
+++ b/src/app/labeling/components/labeling.component.ts
@@ -775,9 +775,13 @@ export class LabelingComponent implements OnInit, OnDestroy {
           let overlayData = {
             token: token,
             group: key,
+            tokenStartIdx: md.tokenStartIdx,
+            tokenEndIdx: md.tokenEndIdx,
+            value: md.value,
             isFirst: i == md.tokenStartIdx,
             isLast: i == md.tokenEndIdx,
             labelName: md.labelingTaskLabel.name,
+            labelingTaskId: md.labelingTaskLabel.labelingTask.id,
             labelId: md.labelingTaskLabel.id,
             labelDisplay: this.getLabelForDisplay(
               md.labelingTaskLabel.name,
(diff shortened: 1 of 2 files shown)


src/app/labeling/components/labeling.component.ts
```diff
@@ -1389,7 +1393,31 @@ export class LabelingComponent implements OnInit, OnDestroy {
     }
   }
 
+  addLabelToMarkedTextFromSuggestion(overlayData) {
```
JWittmeyer, line 1396: no checks are made for the given weak supervision label. This means that the user could accept the same label multiple times or create an overlapping issue with other manual labels (or itself). 
image

jhoetter:
check [gateway](https://github.com/code-kern-ai/refinery-gateway/pull/46) and [submodule](https://github.com/code-kern-ai/refinery-submodule-model/pull/11)

src/app/labeling/components/labeling.component.ts
```diff
@@ -1389,7 +1393,31 @@ export class LabelingComponent implements OnInit, OnDestroy {
...
+    const dataEntry = {
+      startIdx: overlayData.tokenStartIdx,
+      endIdx: overlayData.tokenEndIdx,
+      value: overlayData.value,
+    };
+
+    this.somethingLoading = true;
```
FelixKirsch, line 1403: when clicking on a weak supervision label, which has an overlapping manual label is not set as expected. But the cursor has a loading spinner attached to it, which isn't removed.

FelixKirsch (approved):

src/app/labeling/components/labeling.component.ts
jhoetter: saw this too, i will create an issue for this. but only saw this in minor cases

PR #60, opened 2022-09-19
adds confidence distribution chart

jhoetter:
use [gateway](https://github.com/code-kern-ai/refinery-gateway/pull/45) and [submodule](https://github.com/code-kern-ai/refinery-submodule-model/pull/10) from this branch.

When you jump into the clickbait project, you should see a confidence distribution chart on the overview page

### DIFF
--- a/src/app/monitor/components/confidence-line-chart/confidence-line-chart.component.html
+++ b/src/app/monitor/components/confidence-line-chart/confidence-line-chart.component.html
@@ -1,0 +1,2 @@
+<kern-line-chart [data]="dataInput">
+</kern-line-chart>
--- a/src/app/project-overview/components/project-overview/project-overview.component.html
+++ b/src/app/project-overview/components/project-overview/project-overview.component.html
@@ -220,7 +220,7 @@
         </div>
 
         <div class="mt-8 w-2/4 grid w-full"
-            [ngClass]="displayGraphsForm.value == DisplayGraphsType.ALL || displayGraphsForm.value == DisplayGraphsType.LABEL_DISTRIBUTION?null:'hidden'">
+            [ngClass]="displayGraphsForm.value == DisplayGraphsType.ALL || displayGraphsForm.value == DisplayGraphsType.CONFIDENCE_DISTRIBUTION?null:'hidden'">
             <div class="text-lg leading-6 text-gray-900 font-medium inline-block">
                 Confidence distribution
             </div>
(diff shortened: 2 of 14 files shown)


src/app/monitor/components/confidence-line-chart/confidence-line-chart.component.html
```diff
@@ -0,0 +1,2 @@
+<kern-line-chart [data]="dataInput">
+</kern-line-chart>
```
JWittmeyer, line 2: why is this wrapped in another component? i know we have it for other charts as well but it doesn't make sense to me

src/app/project-overview/components/project-overview/project-overview.component.html
```diff
@@ -219,6 +219,22 @@
             </div>
         </div>
 
+        <div class="mt-8 w-2/4 grid w-full"
+            [ngClass]="displayGraphsForm.value == DisplayGraphsType.ALL || displayGraphsForm.value == DisplayGraphsType.LABEL_DISTRIBUTION?null:'hidden'">
```
JWittmeyer: is this meant to be visible with the LABEL_DISTRIBUTION if you select it or is another state missing here?

src/app/monitor/components/confidence-line-chart/confidence-line-chart.component.html
jhoetter: I only got it working that way tbh haha. Thought this was required for the modules.

src/app/project-overview/components/project-overview/project-overview.component.html
jhoetter: no, you're right, i missed that. will add it

JWittmeyer:
Something seems to be up with the percentile display:
image
this happens when I use the no sample number

- [x] resolved

jhoetter:
discussed the point re 100 samples, so I resolved it :)

FelixKirsch:
In some settings, taking the first 100 records for the confidence chart results in a bad representation of the project. I played around with the conversational-ai project and deleted one heuristic after another and triggered new weak supervision runs. At some point, the confidence chart showed around 100% confidence for all records, while there were still a significant amount of labels with lower weak supervision confidence. I send you the project export.

JWittmeyer:
if the selected amount is < 100 (e.g. with a slice filter) this will resulting < 100 percentil

- [x] resolved

JWittmeyer (approved):

PR #66, opened 2022-09-28
Release 1.3.4 fixes

lumburovskalina:
- https://github.com/code-kern-ai/refinery-ui/pull/66
- https://github.com/code-kern-ai/refinery-gateway/pull/55
- https://github.com/code-kern-ai/refinery-submodule-model/pull/17

### DIFF
--- a/src/app/base/components/heuristic-statuses/heuristic-statuses.component.ts
+++ b/src/app/base/components/heuristic-statuses/heuristic-statuses.component.ts
@@ -1,4 +1,5 @@
 import { Component, Input, OnChanges } from '@angular/core';
+import { Router } from '@angular/router';
 
 @Component({
   selector: 'kern-heuristic-statuses',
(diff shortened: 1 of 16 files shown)


src/app/base/components/heuristic-statuses/heuristic-statuses.component.ts
```diff
@@ -14,8 +15,11 @@ export class HeuristicStatusesComponent implements OnChanges {
   dataTip: string;
   statusName: string;
   color: string;
+  isAttributesOrSettingsPage: boolean;
 
-  constructor() { }
+  constructor(private router: Router) {
+    this.isAttributesOrSettingsPage = this.router.url.indexOf('attributes') > -1 || this.router.url.indexOf('settings') > -1 ? true : false;
```
JWittmeyer: this is decided on the router url. I think we might want to change it to an input e.g. initialCaption that is 'Initial' by default and can be overwritten. 
Depending on an URL in a component feels a bit unreliable

JWittmeyer (approved):

Issue #72, opened 2022-10-13
Modernize Comment Feature


JWittmeyer:
- gives more context of what you want to comment on. A modal mostly blocks the screen, whereas a slideover gives you full access of the actual page
- since the user isn't yet with an image the differentiation will be harder so a marker for "own comment" & a hover on the image with the email 

Example:
<img width="853" alt="image (1)" src="https://user-images.githubusercontent.com/91723236/195526197-589100d2-4840-4ebf-8411-bf75ffdba3f3.png">

lumburovskalina:
Steps to implement the feature:
- Check Tailwind UI existing components
- Build the template with Tailwind UI according to the design mockup
- Add the implemented functionalities from the previous version
- Test and code review

PR #73, opened 2022-10-13
Modernize Comment Feature

lumburovskalina:
Steps to implement redesign/modernization of comment feature:
1. Check Tailwind UI existing components
2. Build the template with Tailwind UI according to the design mockup
3. Add the implemented functionalities from the previous version
4. Test and code review

Issue:
https://github.com/code-kern-ai/refinery-ui/issues/72 

PRs: 
1. https://github.com/code-kern-ai/refinery-ui/pull/73
2. https://github.com/code-kern-ai/refinery-submodule-model/pull/19
3. https://github.com/code-kern-ai/refinery-gateway/pull/70


JWittmeyer:
had to `npm install` and push changes for package log changes and not being able to start -> please double check if it's still working on your side :)

JWittmeyer:
different branch name than the others :)

src/app/base/components/comment/comment-helper.ts
```diff
@@ -340,6 +346,10 @@ export class CommentDataManager {
     private buildCurrentDataOrder() {
         this.currentDataOrder = [];
         for (var key in this.currentData) {
+            const findCurrentUser = this.allUsers.find(u => u.id == this.currentData[key].created_by);
+            const avatarSelector = (findCurrentUser.firstName[0].charCodeAt(0) + findCurrentUser.lastName[0].charCodeAt(0)) % 5;
+            const avatarUri = "assets/avatars/" + avatarSelector + ".png"
```
JWittmeyer: i think these two lines are used a lot in the app -> maybe we should build a helper function that gets user and returns the URI so in the future when we want to change it we don't need to change it everywhere :)

src/app/base/components/comment/comment-helper.ts
```diff
@@ -340,6 +346,10 @@ export class CommentDataManager {
     private buildCurrentDataOrder() {
         this.currentDataOrder = [];
         for (var key in this.currentData) {
+            const findCurrentUser = this.allUsers.find(u => u.id == this.currentData[key].created_by);
```
JWittmeyer, line 350: what happens if the user isn't present? (e.g. deletion or import)
(the mention uri function could handle null users with a default :)

src/app/base/components/comment/comment.component.ts
```diff
@@ -21,13 +39,16 @@ export class CommentComponent implements OnInit, OnDestroy {
   }
   commentIdOptions: any[];
   allOpen: boolean = false;
+  isSlideOverOpen: boolean = false;
+  positionComment: string = 'right';
```
JWittmeyer: let's create an enum for these in the comment-helper (
RIGHT = "RIGHT",
LEFT = "LEFT"

src/app/base/components/comment/comment.component.ts
```diff
@@ -21,13 +39,16 @@ export class CommentComponent implements OnInit, OnDestroy {
...
 
   constructor() { }
   ngOnDestroy(): void {
   }
   ngOnInit(): void {
     this.initDataManger();
     UserManager.registerAfterInitActionOrRun(this, () => this.initUsers(), true);
+    localStorage.setItem('commentPosition', this.positionComment);
```
JWittmeyer: i think this isn't what we want (on init I would assume we need to load it from the local storage and then set it to the variable if already present)

src/app/base/components/dropdown/dropdown.component.html
```diff
@@ -5,6 +5,7 @@
         <button #dropdownOpenButton type="button" *ngIf="dropdownOptions.buttonVersion == 'default'"
             class="inline-flex rounded-md border shadow-sm px-4 items-center text-xs font-semibold cursor-pointer focus:ring-offset-2 focus:ring-offset-gray-400"
             id="menu-button" aria-expanded="true" aria-haspopup="true" [disabled]="dropdownOptions.isDisabled"
+            [ngStyle]="dropdownOptions.addScrollXDropdownOptions ? { 'max-width': '192px' } : {}"
```
JWittmeyer: this is a very fixed value for a very unfixed issue. either we need another option for it or at least an info in the addScrollXDropdownOptions to know that it's set to 192px max width

src/app/base/components/dropdown/dropdown.component.html
```diff
@@ -37,9 +38,10 @@
     </a>
 
     <div #dropdownOptionsDiv
-        class="origin-top-right absolute mt-2 rounded-md shadow-lg bg-white ring-1 ring-black ring-opacity-5 focus:outline-none hidden"
-        [ngClass]="[dropdownClassList, dropdownOptions.isButtonSampleProjects ? 'width-sample-projects' : '', dropdownOptions.buttonVersion != '...' ? 'min-w-full': '']"
-        role="menu" aria-orientation="vertical" aria-labelledby="menu-button" tabindex="-1">
+        class="origin-top-right absolute mt-2 rounded-md shadow-lg bg-white ring-1 ring-black ring-opacity-5 focus:outline-none hidden overflow-y-auto"
+        [ngClass]="[dropdownClassList, dropdownOptions.isButtonSampleProjects ? 'width-sample-projects' : '', dropdownOptions.buttonVersion != '...' ? 'min-w-full': '', dropdownOptions.addScrollXDropdownOptions ? 'overflow-x-scroll w-24' : '']"
+        [ngStyle]="dropdownOptions.addScrollXDropdownOptions ? { 'max-height': '200px' } : {}" role="menu"
```
JWittmeyer: why is an x limitation changing the max height attribute?

JWittmeyer:
comments always start right on reload (see local storage code comment)

- [x] resolved

JWittmeyer:
i think we should give the -> a tooltip as well since it might not be obvious ('Move comment window right' & 'Move comment window left')
image

- [x] resolved

JWittmeyer:
scrollbar always visible even if it's not necessary:
image

- [x] resolved

JWittmeyer:
private comment has no pointer changes and text click should also change the value
image

- [x] resolved


____________________________

checkbox still missing cursor pointer

JWittmeyer:
class leading-4 doesn't take effect in text areas (should be line-height:1rem)

- [x] resolved

JWittmeyer:
i think we should center the heading (comments) to clearly indicate a difference between the button <- and the heading
image

- [x] resolved

JWittmeyer:
i think we should remove the no text provided from the console.log (in dropdown component) to prevent irritation
image

- [x] resolved

JWittmeyer:
Comment block "jumps" when switching from no comments to comments

demo

- [x] resolved

JWittmeyer:
Tooltips are cut of on right side
image

- [x] resolved

JWittmeyer:
cant click on sidebar after closing the comment on right side
demo

- [x] resolved

src/app/util/helper-functions.ts
```diff
@@ -73,4 +73,9 @@ export function parseLogData(logs: string[], isType: InformationSourceType = nul
             wrapper.substr(wrapper.indexOf(' ') + 1)
         );
     });
+}
+
+export function getUserAvatarUri(user) {
+    const avatarSelector = user != undefined ? (user.firstName[0].charCodeAt(0) + user.lastName[0].charCodeAt(0)) % 5 : 0;
+    return "assets/avatars/" + avatarSelector + ".png";
```
JWittmeyer: with this approach, only an undefined user will result in the default 0.
Might run into error for a user not present in the sqlliste db (kratos) but in the db (imported users)

```suggestion
    let avatarId = 0;
    if (user && user.firstName && user.lastName){
        avatarId = user.firstName[0].charCodeAt(0) + user.lastName[0].charCodeAt(0)) % 5;
    }
    return "assets/avatars/" + avatarSelector + ".png";
```

src/app/base/components/dropdown/dropdown.component.html
```diff
@@ -5,6 +5,7 @@
         <button #dropdownOpenButton type="button" *ngIf="dropdownOptions.buttonVersion == 'default'"
             class="inline-flex rounded-md border shadow-sm px-4 items-center text-xs font-semibold cursor-pointer focus:ring-offset-2 focus:ring-offset-gray-400"
             id="menu-button" aria-expanded="true" aria-haspopup="true" [disabled]="dropdownOptions.isDisabled"
+            [ngStyle]="dropdownOptions.maxWidth? { 'max-width': dropdownOptions.maxWidth } : {}"
```
JWittmeyer, line 8: the width of the options should also be applied to the options (at least if the comment says so)

JWittmeyer:
Not sure why we need a scrollbar (and max-width) when there is still space left.
See comment above
image

- [x] resolved

JWittmeyer (approved):

PR #75, opened 2022-10-17
Enable Attribute calculation for different data types

lumburovskalina:
implements:
- https://github.com/code-kern-ai/refinery/issues/142

works with:
- https://github.com/code-kern-ai/refinery-gateway/pull/68
- https://github.com/code-kern-ai/refinery-ac-exec-env/pull/6
- https://github.com/code-kern-ai/refinery-ui/pull/75
- https://github.com/code-kern-ai/refinery-tokenizer/pull/11
- https://github.com/code-kern-ai/refinery-submodule-model/pull/21

## Frontend
Tasks:
- Add a new kern-dropdown with all data types on the attribute-calculation page
- Set conditions when it should be enabled/disabled
- Change datatype on project-settings (remove dropdown and add it as a text)
- Implement modal for adding a new attribute with name and data type
- Update queries and mutation according to the new modal
- Visual and UI improvements (small issues to keep the consistency)
- Testing and review 
Estimation time FE: 9h

### DIFF
--- a/src/app/base/components/dropdown/dropdown-helper.ts
+++ b/src/app/base/components/dropdown/dropdown-helper.ts
@@ -31,6 +31,7 @@ import { FormArray } from "@angular/forms";
  * @addScrollXDropdownOptions {boolean, optional} - checks if the dropdown options should have a scrollX for longer texts with limitted space
  * @maxHeight {string, optional} - sets the max height of the dropdown options
  * @maxWidth {string, optional} - sets the max width of the dropdown options
+ * @hasFullWidth {string, optional} - sets the width of the dropdown options and the button to 100%
 */
 export type DropdownOptions = {
     optionArray: string[] | FormArray[] | any[];
@@ -62,5 +63,6 @@ export type DropdownOptions = {
     addScrollXDropdownOptions: boolean;
     maxHeight?: string;
     maxWidth?: string;
+    hasFullWidth?: boolean;
 };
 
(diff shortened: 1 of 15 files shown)


JWittmeyer:
Uses somewhat bold font for tooltips (card-title class changes font weight)
image

- [x] resolved

JWittmeyer:
i think since the modal is not located "outside" the hover effect on the button is always visible once the modal is open
image

- [x] resolved

JWittmeyer:
different styling for attribute calcuclation modal:
Example Embeddings:
image

Attributes:
image

Noticed differences:
- [x] different base width (labeling task windows is also bigger - they should all use a similar size (max-w-xs)) 
- [x] dropdown not as wide as other element
- [x] no helping text ("Choose a name for your attribute and pick a datatype you want to use")
- [x] Button Caption is only Create - I think this should be Add Attribute

JWittmeyer:
Default for different datatypes doesn't match the datatype.
image

- [x] resolved

JWittmeyer:
I think since we now introduce different datatype we need to make clear what datatype an attribute is on different pages as well.
image

Let's add the datatype as color coded information to the exec envs and as tooltip (e.g. Integer - Click to copy)
Color coding proposition (Johannes has final say):
- Category -> amber
- Text -> lime
- Boolean -> cyan
- Integer -> indigo
- Float -> purple

Also for labeling page a data tip e.g. "Datatype: Integer"

- [x] resolved

JWittmeyer (approved):

PR #77, opened 2022-10-21
Empty fields when uploading records with new attributes fix on data b…

lumburovskalina:
Steps to fix bug:

1. Upload the additional record to reproduce bug.  (15minutes)
2. Set condition in the data-browser and labeling page when the attribute is not present (30minutes)
3. If the condition is true, set a marker (maybe a warning icon)? (15minutes)
4. Testing and review (1h)

Estimation time: 2h

### DIFF
--- a/src/app/labeling/components/labeling.component.html
+++ b/src/app/labeling/components/labeling.component.html
@@ -132,8 +132,24 @@
                                         </label>
                                     </div>
                                     <div class="col-span-3 self-start">
-                                        <ng-template ngFor let-task [ngForOf]="fullRecordData.tasks[order.key]"
-                                            let-taskIndex="index">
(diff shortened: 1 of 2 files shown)


src/app/labeling/components/labeling.component.html
```diff
@@ -132,8 +132,24 @@ <h1 class="text-lg text-gray-900 mb-2">Warning
                                         </label>
                                     </div>
                                     <div class="col-span-3 self-start">
-                                        <ng-template ngFor let-task [ngForOf]="fullRecordData.tasks[order.key]"
-                                            let-taskIndex="index">
+                                        <span *ngIf="fullRecordData.tasks[order.key] == null;else displayData"
```
JWittmeyer: i think we also need to check for empty strings :)

JWittmeyer (approved):

PR #80, opened 2022-10-31
Reduce compilation time

lumburovskalina:
Issue: 
https://github.com/code-kern-ai/refinery/issues/152 

## Compilation time
Tasks:
- Since version Angular 9, the AOT compiler is used (in the angular.json for the build process we should set this to true)
- Additionally, in the tsconfig.json we should enable the Ivy compiler
- Additional configurations in angular.json (buildOptimization should be set to false and replaced with other options since it is taking too much time) (sourceMap should be also set to true since it is used only for debugging) (vendorChunk, extractLicenses, namedChunks should be set to false)
- Increase Node's memory limit
- browserTarget should be changed to :build:development (changes from Angular 13)
- defaultConfiguration should be set to development

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -71,6 +71,7 @@
               "progress": false,
               "vendorChunk": true,
               "buildOptimizer": false,
+              "subresourceIntegrity": false,
               "budgets": [
                 {
                   "type": "initial",
@@ -83,6 +84,18 @@
                   "maximumError": "4kb"
                 }
               ]
+            },
+            "development": {
+              "optimization": false,
+              "outputHashing": "none",
+              "aot": true,
+              "extractLicenses": false,
+              "statsJson": false,
+              "progress": true,
+              "vendorChunk": true,
+              "buildOptimizer": false,
+              "sourceMap": true,
+              "namedChunks": false
             }
           }
         },
--- a/angular.json
+++ b/angular.json
@@ -60,13 +60,17 @@
                   "with": "src/environments/environment.prod.ts"
                 }
               ],
-              "optimization": true,
-              "outputHashing": "all",
+              "optimization": false,
+              "outputHashing": "none",
               "sourceMap": false,
               "namedChunks": false,
-              "extractLicenses": true,
-              "vendorChunk": false,
-              "buildOptimizer": true,
+              "showCircularDependencies": false,
+              "aot": true,
+              "extractLicenses": false,
+              "statsJson": false,
+              "progress": false,
+              "vendorChunk": true,
+              "buildOptimizer": false,
               "budgets": [
                 {
                   "type": "initial",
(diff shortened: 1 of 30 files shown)


angular.json
```diff
@@ -60,18 +60,47 @@
...
+              "progress": false,
               "vendorChunk": false,
-              "buildOptimizer": true,
+              "buildOptimizer": false,
+              "subresourceIntegrity": false,
               "budgets": [
                 {
                   "type": "initial",
-                  "maximumWarning": "5mb",
-                  "maximumError": "8mb"
```
JWittmeyer: not sure we should change these -> only the local build should be affected by the changes, not the drone build so the difference in maximum build size doesn't make sense imo

package.json
```diff
@@ -3,10 +3,10 @@
   "version": "0.0.0",
   "scripts": {
     "ng": "ng",
-    "start": "ng serve",
+    "start": "ng serve --configuration development",
     "start-proxy": "ng serve --proxy-config proxy.conf.json",
     "build": "ng build",
-    "build-prod": "ng build --prod --aot --build-optimizer --outputPath=./dist/out",
```
JWittmeyer, line 9: see above, for production (drone builds) this shouldn't be changed, right?

lumburovskalina: yes, that is correct, I reverted those changes

JWittmeyer:
Optimization looks good, i have improvements over all checked aspects:
start normal ~ 2m 31 sec
start enhanced ~ 3m 20 sec (had some initial install tasks)
start enhanced 2nd ~ 1m 22 sec
change normal html ~ 1m 22 sec
change enhanced html ~ 28 sec
change normal ts ~ 1m 16sec
change enhanced ts ~ 25 sec

JWittmeyer:
Error on projects page:
image

- [x] resolved

JWittmeyer:
Error databrowser:
image

- [x] resolved

JWittmeyer:
Pretty much all pages have either errors or warnings.
Example error Warnings:
image

- [x] resolved


They seem to be related to existing code only noticeable after the change.

JWittmeyer:
Filtering on embedding creation doesnt work
image

- [x] resolved

JWittmeyer:
Crowd labeler:
image

- [x] resolved

JWittmeyer:
Not sure if related but felix mentioned this in the last test as well:
image

Steps to reproduce:
1. Import Clickbait sample
2. create labelingtask on running_id 
3. switch to labeling (data browser looks fine)

- [x] resolved

JWittmeyer:
filtering in embeddings still doesn't work. (no console error but the function doesn't work anymore)

- [x] resolved

JWittmeyer:
crowd heuristic still doesn't work
image

- [x] resolved

JWittmeyer (approved):

Issue #62, opened 2022-09-26
Add a mandatory confirmation to the deletion of a filter in the Data Browser


DerEchteFeuerpfeil:
This could be a modal just like we have for embedding deletion etc..

JWittmeyer:
personally not a fan of this since I don't like the extra x clicks. Maybe on hitting clear the filter is stored locally and can be "reset" by clicking on the clear button again?

DerEchteFeuerpfeil:
@JWittmeyer 
Are we talking about the same thing? I mean this deletion here:
![image](https://user-images.githubusercontent.com/57487741/203324870-b81d340e-6c75-4c0a-b27b-2f4b9a1863e0.png)

It could happen that I click on this by accident and then the filter is gone without a warning. If it was by accident I don't even know which filter is missing then since there is no notification or anything.

JWittmeyer:
oh :D sorry i thought you  meant the clear filter button
![image](https://user-images.githubusercontent.com/91723236/203326203-db8fd3d6-61dc-4024-881c-28b1e7964f17.png)

for that no problem :)

lumburovskalina:
This issue is implemented as part of [refactoring modals with shared component](https://github.com/code-kern-ai/refinery-ui/pull/89). The estimation and steps to implement are included in that issue.

PR #89, opened 2022-11-22
Modals refactor with a shared component

lumburovskalina:
Implements:
- https://github.com/code-kern-ai/refinery/issues/187 
- https://github.com/code-kern-ai/refinery-ui/issues/62

Replacing all modals with the shared component kern-modal and adding delete confirmation modal for data slices.

### DIFF
--- a/src/app/data/components/data-browser/data-browser.component.html
+++ b/src/app/data/components/data-browser/data-browser.component.html
@@ -1221,12 +1221,12 @@
 
 <kern-modal [isOpen]="dataBrowserModals.deleteSlice.open" closeButton="X" [modalBoxStyle]="{'width':'30rem'}"
     [abortButton]="{
-    buttonCaption: 'Delete',
+    buttonCaption: 'Delete data slice',
     emitObject: this, emitFunction: deleteSlice
 }" (optionClicked)="dataBrowserModals.deleteSlice.open = false">
-    <h1 class="text-lg text-gray-900 mb-2 text-center">Delete data slice</h1>
+    <h1 class="text-lg text-gray-900 mb-2 text-center">Warning</h1>
     <div class="text-sm text-gray-500 my-2 text-center">
-        Are you sure you want to delete selected slice?
+        Are you sure you want to delete this data slice?
     </div>
 </kern-modal>
 
--- a/src/app/data/components/data-browser/data-browser.component.html
+++ b/src/app/data/components/data-browser/data-browser.component.html
@@ -41,76 +41,6 @@
                         (ngModelChange)="filterAvailableSlices()" class="w-full input input-sm input-bordered p-2">
 
                 </div>
-                <input #saveInputCheckbox type="checkbox" id="save-slice-modal" class="modal-toggle"
-                    (change)="focusModalInputBox($event,'dataSliceName')">
-                <div class="modal">
-                    <div class="modal-box text-black bg-white justify-center">
-                        <div class="flex flex-grow justify-center text-lg leading-6 text-gray-900 font-medium">
-                            Filter creation</div>
-                        <div class="mb-2 flex flex-grow justify-center text-sm">
-                            Save or update a filter for the data browser.</div>
-                        <p class="pt-4">You have two choices.
-                            <label class="block font-bold mt-2">Dynamic:</label>
-                            Stores the current filter critera and is reapplied whenever the slice is
-                            selected.<br><br>
-                            <label class="block font-bold mt-2 ">Static:</label>
-                            Used for visualization filters and scope of
-                            work for annotators & domain experts.
-                            Stores the current records provided by your filter.
-                            These are limited to 10,000 records and can be refreshed if nessecary.
-                        </p>
-                        <div class="mt-2">
-                            <div class="flex">
-                                <div class="inline-flex">
-                                    <div class="form-control">
-                                        <label class="cursor-pointer label">
-                                            <span class="label-text mr-1">Dynamic</span>
-                                            <input #inputDynamic type="radio"
-                                                style="--focus-shadow:none;accent-color:rgb(29,78,216)"
-                                                (click)="setStatic(inputDynamic,inputStatic,false)"
-                                                class="h-4 w-4 border-gray-200 cursor-pointer" checked [value]="true">
-                                        </label>
-                                    </div>
-                                    <div class="form-control">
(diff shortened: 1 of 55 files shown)


src/app/data/components/data-browser/data-browser.component.ts
```diff
@@ -2088,4 +2091,20 @@ export class DataBrowserComponent implements OnInit, OnDestroy {
...
+  setEmbeddingId(selectedValue: string) {
+    this.dataBrowserModals.findOutliers.embeddingId = this.similarSearchHelper.embeddings[selectedValue].id;
+  }
+
+  setEmbeddingIdSS(selectedValue: string) {
+    this.dataBrowserModals.similaritySearch.embeddingId = this.similarSearchHelper.embeddings[selectedValue].id;
+  }
 }
+function createDataBrowserModals(): DataBrowserModals {
+  throw new Error('Function not implemented.');
```
JWittmeyer: this doesn't seem right :D

src/app/data/components/data-browser/data-browser.component.html
```diff
@@ -1288,4 +1004,241 @@ <h1 class="text-lg text-gray-900 mb-2">Info</h1>
...
+                        <path fill="#E8CD20" stroke-linecap="round" stroke-linejoin="round"
+                            d="M12 9v2m0 4h.01m-6.938 4h13.856c1.54 0 2.502-1.667 1.732-3L13.732 4c-.77-1.333-2.694-1.333-3.464 0L3.34 16c-.77 1.333.192 3 1.732 3z" />
+                    </svg>
+                </div>
+            </div>
+        </div>
+        <div class="flex w-full justify-between mt-2">
+            <input #inputSliceName type="text" class="w-1/2 input input-bordered input-sm" placeholder="Enter name"
+                name="dataSliceName" [value]="dataBrowserModals.filter.name"
+                (input)="setFilterName(inputSliceName.value)" (keyup)="checkNameExists(inputSliceName.value)"
```
JWittmeyer: not quite sure why these are in different methods both should work in keyup.
Also setting the name should only be necessary right before accepting, right?

src/app/data/components/data-browser/data-browser.component.html
```diff
@@ -1288,4 +1004,241 @@ <h1 class="text-lg text-gray-900 mb-2">Info</h1>
...
+</kern-modal>
+
+<kern-modal [isOpen]="dataBrowserModals.sliceInfo.open" closeButton="X"
+    (optionClicked)="dataBrowserModals.sliceInfo.open = false">
+    <div class="flex flex-grow justify-center mb-4 font-bold">Slice information</div>
+    <div *ngFor="let item of dataBrowserModals.sliceInfo.data | keyvalue : orderOriginal">
+        <ng-template [ngIf]="item.key == 'Link'" [ngIfElse]="normalDisplay">
+            <div class="mt-3 flex rounded-md">
+                <span
+                    class="inline-flex items-center rounded-l-md border border-r-0 border-gray-300 bg-gray-50 px-3 text-gray-500 sm:text-sm">{{item.value.startsWith("https")?
```
JWittmeyer, line 1141: function call inside html -> i think it was there before the change but since I am noticing some performance issues I think we need to remove as many as possible :)

src/app/model-callbacks/components/model-callbacks.component.ts
```diff
@@ -33,16 +34,16 @@ export class ModelCallbackComponent implements OnInit, OnDestroy {
   labelingTasksQuery$: any;
 
   modalOpen: boolean = false;
-  selectedInformationSources = [];
+  // selectedInformationSources = [];
```
JWittmeyer: comment

src/app/model-callbacks/components/model-callbacks.component.ts
```diff
@@ -33,16 +34,16 @@ export class ModelCallbackComponent implements OnInit, OnDestroy {
...
   informationSourcesArray = [];
   filteredSourcesList = [];
 
   openTab: number = -1;
   functionName: string = '';
   labelingTaskId: any;
   description: string;
-  selectionList: string = "";
-  @ViewChild("deleteSelectedHeuristics") deleteSelectedHeuristics: ElementRef;
+  // selectionList: string = "";
```
JWittmeyer: comment

JWittmeyer:
Probably not from this pr but i noticed the checkbox for databrowser configurations aren't clickable via the text & don't have cursor pointer
image

- [x] resolved

JWittmeyer:
in the previsou versoin the save slice name was in the line of the action line. Since this isn't the case anymore i'd make the modal a little smaller (width35rem) and give the input box the full width
image

- [x] resolved

JWittmeyer:
neither close nor delete could be clicked (and cursor was not allowed)

image

(only occurred in clickbait with one lf selected -> text hove had issues the actual button not? -> in general a bit weird case :D )

- [x] discussed
- [x] resolved

src/app/data/components/data-browser/data-browser.component.html
lumburovskalina: Yes, that is true, both should be in keyup. 
Also yes for the second thing, I will move the logic from setFilterName to both methods for update and create slice.

JWittmeyer:
Deleting an heuristic on the details page doesn't seem to redirect directly (the alert shouldn't be shown)

- [x] discussed
- [x] resolved

JWittmeyer:
crowd heuristic delete button possitioned different than others.
image

either without margin left & some marging top or in the same line

- [x] resolved

src/app/data/components/data-browser/data-browser.component.html
lumburovskalina: Or not, I need the setFilterName as we type so I would check if the value of the input field is empty

JWittmeyer:
stacked modals seem to have some issues. Noticable when opening the weak supervision modal on heuristics in clickbait -> seems that the pointer events none isn't enough. I think this should be changed to hidden to prevent overlap
image

- [x] discussed
- [x] resolved

JWittmeyer:
the kern status display seems to be fixed to have a ml-2 somewhere. This causes the modal for weak supervision to be unaligned
image

I'd solve this with an input for a class that defaults to ml-2 and can be overwritten or something 

- [x] discussed
- [x] resovled

JWittmeyer (approved):

PR #92, opened 2022-11-29
Adds bricks integrator

### LINKED ISSUE #186: [BUG] - Highlighting in databrowser ignores case flag
**Describe the bug**
The new addition to the data browser for case-sensitive filtering work however the highlighting still highlights case insensitive

**To Reproduce**
1. Import Clickbait sample project
2. Click on data browser
3. Create a filter for EU on headline & case sensitive flag
4. running_id 13 will highlight two instances one with EU and one with Eu. 

**Expected behavior**
Only EU should be highlighted

**Screenshots**
image

 - OS: all
 - Browser all
 - Version 1.6.0

**Additional context**
The filter function isn't broken, only the highlighting

### LINKED ISSUE #190: [BUG] - Highlighting in data-browser causes performance issues
**Describe the bug**
When using the highlight function in the data browser the overall performance degrades. E.g. a modal that needs to be opened to save a filter takes 2-3 seconds.

**To Reproduce**
1. Go to data browser
2. Add filter with highlighting (attribute filter)
3. try to open a modal
4. wait

**Expected behavior**
faster response time

 - OS: Windows
 - Browser Chrome
 - Version 1.6

**Additional context**
Looking into the dom tree shows that the constant redraw of the highlight module seems to be the culprit

JWittmeyer:
- Implements https://github.com/code-kern-ai/refinery/issues/185
- Fixes https://github.com/code-kern-ai/refinery/issues/186
- Fixes https://github.com/code-kern-ai/refinery/issues/190

also replaces highlight module with own


lumburovskalina:
Suggestion: since we have this.http.post and this.http.get, maybe we can add the requests in a separate service file?

- [x] discussed

src/app/base/components/bricks-integrator/helper/type-helper.ts
```diff
@@ -0,0 +1,164 @@
...
+        type: null,
+        comment: null,
+        optional: false,
+        options: {}
+    }
+}
+
+export function bricksVariableNeedsTaskId(variableType: BricksVariableType): boolean {
+    switch (variableType) {
+        case BricksVariableType.EMBEDDING:
```
FelixKirsch, line 157: does this case have any effect?

JWittmeyer: yes, in javascript if there is no break they "fall through" so this means both embedding and label use the true return

FelixKirsch (approved):

src/app/base/components/bricks-integrator/helper/type-helper.ts
FelixKirsch: ah, didn't remember that

PR #93, opened 2022-12-01
Add logout option for demo user

lumburovskalina:
Implements:
- https://github.com/code-kern-ai/refinery/issues/188

PRs:
- https://github.com/code-kern-ai/refinery-ui/pull/93

### DIFF
--- a/src/app/base/components/header/header.component.html
+++ b/src/app/base/components/header/header.component.html
@@ -1,7 +1,7 @@
 <header class="sticky top-0 z-50 w-full">
     <div
         class="relative z-10 flex-shrink-0 h-16 bg-white border-b border-gray-200 shadow-sm flex justify-between items-center">
-        <div *ngIf="hideLogout"
+        <div *ngIf="isDemo"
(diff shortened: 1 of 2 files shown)


src/app/base/components/header/header.component.html
```diff
@@ -152,6 +152,13 @@
                         }" (optionClicked)="executeOption($event)"></kern-dropdown>
                 </div>
             </div>
+            <div class="flex justify-center overflow-visible" *ngIf="hideLogout">
```
JWittmeyer: not sure if this makes sense regarding the naming of the variable "hideLogout" -> if it's true we give a logout? 

i'd go without adding a new block but changing the existing one with currently *ngIf="!hideLogout"


optionArray: isDemo? ['Logout']: showConfigSettings ? ['Account Settings', 'Change config', 'Logout'] : ['Account Settings', 'Logout'],

and create a variable isDemo to access and change the hidelogout to work with a different condition (currently fixed to false)

JWittmeyer (approved):

PR #94, opened 2022-12-02
Modal fixes after merging refactoring

lumburovskalina:
Modals fixes after merging refactoring:
- Tooltip on last weak supervision run makes the modal scrollable.
- Focus on input in modals for adding labels, labeling tasks and attributes is missing.

### DIFF
--- a/src/app/projects/components/project-settings/project-settings.component.ts
+++ b/src/app/projects/components/project-settings/project-settings.component.ts
@@ -555,18 +555,12 @@ export class ProjectSettingsComponent implements OnInit, OnDestroy, AfterViewIni
 
   focusModalInputBox(event: Event, inputBoxName: string) {
     event.preventDefault();
-    const inputChildren = document.getElementsByName(inputBoxName);
-    if (inputChildren.length > 0) {
-      const node = inputChildren[0];
-      if (
-        node instanceof HTMLElement &&
-        node.getAttribute('name') == inputBoxName
-      ) {
-        setTimeout(() => {
-          node.focus();
-        }, 0);
-        return;
-      }
+    const inputChildren = document.getElementById(inputBoxName);
+    if (inputChildren && inputChildren instanceof HTMLElement && inputChildren.getAttribute('name') == inputBoxName) {
+      setTimeout(() => {
+        inputChildren.focus();
+      }, 0);
+      return;
     }
   }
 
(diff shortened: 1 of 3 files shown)


src/app/projects/components/project-settings/project-settings.component.ts
```diff
@@ -554,20 +554,18 @@ export class ProjectSettingsComponent implements OnInit, OnDestroy, AfterViewIni
...
-          if (
-            node instanceof HTMLElement &&
-            node.getAttribute('name') == inputBoxName
-          ) {
-            node.focus();
-            return;
-          }
-        }
+    event.preventDefault();
+    const inputChildren = document.getElementsByName(inputBoxName);
```
JWittmeyer: get elements by name isn't superfast and an array. get element by id should work better for this to work we need to give it an id (and to prevent autocomplete values autocomplete="off" I think)

src/app/projects/components/project-settings/project-settings.component.ts
```diff
@@ -554,20 +554,18 @@ export class ProjectSettingsComponent implements OnInit, OnDestroy, AfterViewIni
...
+    if (inputChildren.length > 0) {
+      const node = inputChildren[0];
+      if (
+        node instanceof HTMLElement &&
+        node.getAttribute('name') == inputBoxName
+      ) {
+        setTimeout(() => {
+          node.focus();
+        }, 0);
+        return;
```
JWittmeyer: with element by id parts of this are unnessecary

src/app/projects/components/project-settings/project-settings.component.ts
```diff
@@ -555,18 +555,12 @@ export class ProjectSettingsComponent implements OnInit, OnDestroy, AfterViewIni
...
-          node.focus();
-        }, 0);
-        return;
-      }
+    const inputChildren = document.getElementById(inputBoxName);
+    if (inputChildren && inputChildren instanceof HTMLElement && inputChildren.getAttribute('name') == inputBoxName) {
+      setTimeout(() => {
+        inputChildren.focus();
+      }, 0);
+      return;
```
JWittmeyer: since we have an id we don't need to check the name so id would go with something like this:

```Typescript

  focusModalInputBox(event: Event, inputBoxId: string) {
    event.preventDefault();
    const inputElement = document.getElementById(inputBoxId) as HTMLInputElement;
    if (inputElement) inputElement.focus();
  }
```

no clue why there is a timeout there, I think we don't need it anymore.

JWittmeyer (approved):

PR #99, opened 2022-12-13
Project settings refactor into subcomponents

lumburovskalina:
Implements:
- https://github.com/code-kern-ai/refinery/issues/192

### DIFF
--- a/src/app/base/components/dropdown/dropdown.component.ts
+++ b/src/app/base/components/dropdown/dropdown.component.ts
@@ -96,7 +96,6 @@ export class DropdownComponent implements OnChanges {
     if (this.dropdownOptions.isOptionDisabled && this.dropdownOptions.isOptionDisabled.length != this.dropdownOptions.optionArray.length) this.hasInputErrors = "array options != isOptionDisabled length\n";
     if (this.dropdownOptions.optionDescriptions && this.dropdownOptions.optionArray.length != this.dropdownOptions.optionDescriptions.length) this.hasInputErrors = "array options != optionDescriptions length\n";
     if (this.dropdownOptions.optionIcons && this.dropdownOptions.optionIcons.length != this.dropdownOptions.optionIcons.length) this.hasInputErrors = "array options != optionIcons length\n";
-    if (this.dropdownOptions.optionTooltips && this.dropdownOptions.optionTooltips.length != this.dropdownOptions.optionArray.length) this.hasInputErrors = "array options != optionTooltip length\n";
     if (this.dropdownOptions.backgroundColors && this.dropdownOptions.backgroundColors.length != this.dropdownOptions.optionArray.length) this.hasInputErrors = "array options != backgroundColors length\n";
 
     if (!this.dropdownOptions.buttonVersion) this.dropdownOptions.buttonVersion = "default";
--- a/src/app/base/components/dropdown/dropdown.component.ts
+++ b/src/app/base/components/dropdown/dropdown.component.ts
@@ -97,7 +97,7 @@ export class DropdownComponent implements OnChanges {
     if (this.dropdownOptions.optionDescriptions && this.dropdownOptions.optionArray.length != this.dropdownOptions.optionDescriptions.length) this.hasInputErrors = "array options != optionDescriptions length\n";
     if (this.dropdownOptions.optionIcons && this.dropdownOptions.optionIcons.length != this.dropdownOptions.optionIcons.length) this.hasInputErrors = "array options != optionIcons length\n";
     if (this.dropdownOptions.backgroundColors && this.dropdownOptions.backgroundColors.length != this.dropdownOptions.optionArray.length) this.hasInputErrors = "array options != backgroundColors length\n";
-
+    if (this.dropdownOptions.optionTooltips && this.dropdownOptions.optionTooltips.length != this.dropdownOptions.optionArray.length) this.hasInputErrors = "array options != optionTooltip length\n";
     if (!this.dropdownOptions.buttonVersion) this.dropdownOptions.buttonVersion = "default";
 
     this.buttonClassList = "";
--- a/src/app/base/entities/project.ts
+++ b/src/app/base/entities/project.ts
@@ -7,4 +7,6 @@ export interface Project {
   projectType: string;
   description?: string;
   numDataScaleUploaded?: number;
+  tokenizer?: string;
+  containsUniqueAttribute?: boolean;
 }
--- a/src/app/base/entities/project.ts
+++ b/src/app/base/entities/project.ts
@@ -8,5 +8,4 @@ export interface Project {
   description?: string;
   numDataScaleUploaded?: number;
   tokenizer?: string;
-  containsUniqueAttribute?: boolean;
 }
--- a/src/app/model-download/components/model-download-component/model-download-component.component.ts
+++ b/src/app/model-download/components/model-download-component/model-download-component.component.ts
@@ -26,7 +26,6 @@ export class ModelDownloadComponentComponent implements OnInit {
   isManaged: boolean = true;
   downloadedModels: any[];
   subscriptions$: Subscription[] = [];
-  lastPage: string;
   loggedInUser: any;
   avatarUri: any;
   downloadedModelsModals: ModelDownloadModals = createDefaultModelDownloadModals();
@@ -43,8 +42,7 @@ export class ModelDownloadComponentComponent implements OnInit {
   ngOnInit(): void {
     this.initForm();
     this.routeService.updateActivatedRoute(this.activatedRoute);
-    this.lastPage = this.activatedRoute.snapshot.queryParamMap.get('lastPage');
-    [this.downloadedModelsQuery$, this.downloadedModelsList$] = this.informationSourceApolloService.getModelProviderInfo();
+    [this.downloadedModelsQuery$, this.downloadedModelsList$] = this.projectApolloService.getModelProviderInfo();
     this.prepareModels();
 
     this.subscriptions$.push(
@@ -83,14 +81,14 @@ export class ModelDownloadComponentComponent implements OnInit {
   }
 
   deleteModel() {
-    this.informationSourceApolloService
+    this.projectApolloService
       .deleteModel(this.downloadedModelsModals.deleteModel.name)
       .pipe(first()).subscribe();
   }
 
   downloadModel() {
     if (this.downloadedModelsModals.addNewModel.form.invalid) return;
-    this.informationSourceApolloService
+    this.projectApolloService
       .downloadModel(this.downloadedModelsModals.addNewModel.form.get('name').value)
       .pipe(first()).subscribe(() => {
         this.downloadedModelsModals.addNewModel.form.reset();
@@ -171,9 +169,6 @@ export class ModelDownloadComponentComponent implements OnInit {
     return findModel !== undefined ? true : false;
   }
 
-  goBack() {
-    this.router.navigate(["../" + this.lastPage], { relativeTo: this.activatedRoute });
-  }
   clickBack() {
     RouteManager.moveBack();
   }
--- a/src/app/projects/components/project-settings/components/data-schema/data-schema.component.ts
+++ b/src/app/projects/components/project-settings/components/data-schema/data-schema.component.ts
@@ -4,7 +4,6 @@ import { Project } from 'src/app/base/entities/project';
 import { ProjectApolloService } from 'src/app/base/services/project/project-apollo.service';
 import { attributeVisibilityStates, getTooltipVisibilityState } from '../../../create-new-attribute/attributes-visibility-helper';
 import { Attribute } from '../../entities/attribute.type';
-import { DataHandlerHelper } from '../../helper/data-handler-helper';
 
 
 @Component({
@@ -12,12 +11,11 @@ import { DataHandlerHelper } from '../../helper/data-handler-helper';
   templateUrl: './data-schema.component.html',
   styleUrls: ['./data-schema.component.scss']
 })
-export class DataSchemaComponent implements OnInit {
+export class DataSchemaComponent implements OnInit, OnChanges {
 
   @Input() project: Project;
   @Input() attributes: Attribute[];
   @Input() pKeyValid: boolean;
-  @Input() dataHandlerHelper: DataHandlerHelper;
 
   tooltipsArray: string[];
   attributeVisibilityStates = attributeVisibilityStates;
@@ -39,23 +37,16 @@ export class DataSchemaComponent implements OnInit {
   }
 
   updateVisibility(option: string, attribute: Attribute) {
-    this.additionalChecksIfChanged(attribute.id);
     const visibility = attributeVisibilityStates[option].value;
     this.projectApolloService.updateAttribute(this.project.id, attribute.id, null, null, null, null, visibility).pipe(first()).subscribe();
   }
 
   updatePrimaryKey(attribute: Attribute) {
-    this.additionalChecksIfChanged(attribute.id);
     this.projectApolloService.updateAttribute(this.project.id, attribute.id, null, !attribute.isPrimaryKey, null, null, null).pipe(first()).subscribe();
   }
 
   private checkStillLoading() {
     this.somethingLoading = this.attributes.length == 0;
-  }
-
-  private additionalChecksIfChanged(attributeId: string) {
-    if (this.dataHandlerHelper.pKeyChanged(this.attributes)) this.dataHandlerHelper.requestPKeyCheck(this.project.id, this.attributes);
-    if (this.dataHandlerHelper.attributeChangedToText(this.attributes)) this.dataHandlerHelper.createAttributeTokenStatistics(this.project.id, attributeId);
   }
 }
 
--- a/src/app/projects/components/project-settings/components/data-schema/data-schema.component.ts
+++ b/src/app/projects/components/project-settings/components/data-schema/data-schema.component.ts
@@ -1,9 +1,10 @@
-import { Component, Input, OnChanges, SimpleChanges } from '@angular/core';
+import { Component, Input, OnChanges, OnInit, SimpleChanges } from '@angular/core';
 import { first } from 'rxjs/operators';
 import { Project } from 'src/app/base/entities/project';
 import { ProjectApolloService } from 'src/app/base/services/project/project-apollo.service';
 import { attributeVisibilityStates, getTooltipVisibilityState } from '../../../create-new-attribute/attributes-visibility-helper';
 import { Attribute } from '../../entities/attribute.type';
+import { DataHandlerHelper } from '../../helper/data-handler-helper';
 
 
 @Component({
@@ -11,40 +12,50 @@ import { Attribute } from '../../entities/attribute.type';
   templateUrl: './data-schema.component.html',
   styleUrls: ['./data-schema.component.scss']
 })
-export class DataSchemaComponent implements OnChanges {
+export class DataSchemaComponent implements OnInit {
 
   @Input() project: Project;
   @Input() attributes: Attribute[];
   @Input() pKeyValid: boolean;
+  @Input() dataHandlerHelper: DataHandlerHelper;
 
   tooltipsArray: string[] = [];
   attributeVisibilityStates = attributeVisibilityStates;
   somethingLoading: boolean = false;
 
-  constructor(private projectApolloService: ProjectApolloService) {
+  constructor(private projectApolloService: ProjectApolloService) { }
+
+  ngOnInit() {
+    this.tooltipsArray = [];
+    this.attributeVisibilityStates.forEach((attribute) => {
+      this.tooltipsArray.push(getTooltipVisibilityState(attribute.value));
+    });
   }
 
   ngOnChanges(changes: SimpleChanges) {
-    this.tooltipsArray = [];
-    this.checkStillLoading();
-    if (changes.attributes) {
-      this.attributeVisibilityStates.forEach((attribute) => {
-        this.tooltipsArray.push(getTooltipVisibilityState(attribute.value));
-      });
+    if (changes.attribute) {
+      this.checkStillLoading();
     }
   }
 
   updateVisibility(option: string, attribute: Attribute) {
+    this.additionalChecksIfChanged(attribute.id);
     const visibility = attributeVisibilityStates[option].value;
     this.projectApolloService.updateAttribute(this.project.id, attribute.id, null, null, null, null, visibility).pipe(first()).subscribe();
   }
 
   updatePrimaryKey(attribute: Attribute) {
+    this.additionalChecksIfChanged(attribute.id);
     this.projectApolloService.updateAttribute(this.project.id, attribute.id, null, !attribute.isPrimaryKey, null, null, null).pipe(first()).subscribe();
   }
 
   private checkStillLoading() {
     this.somethingLoading = this.attributes.length == 0;
+  }
+
+  private additionalChecksIfChanged(attributeId: string) {
+    if (this.dataHandlerHelper.pKeyChanged(this.attributes)) this.dataHandlerHelper.requestPKeyCheck(this.project.id, this.attributes);
+    if (this.dataHandlerHelper.attributeChangedToText(this.attributes)) this.dataHandlerHelper.createAttributeTokenStatistics(this.project.id, attributeId);
   }
 }
 
--- a/src/app/projects/components/project-settings/components/data-schema/data-schema.component.ts
+++ b/src/app/projects/components/project-settings/components/data-schema/data-schema.component.ts
@@ -19,7 +19,7 @@ export class DataSchemaComponent implements OnInit {
   @Input() pKeyValid: boolean;
   @Input() dataHandlerHelper: DataHandlerHelper;
 
-  tooltipsArray: string[] = [];
+  tooltipsArray: string[];
   attributeVisibilityStates = attributeVisibilityStates;
   somethingLoading: boolean = false;
 
--- a/src/app/projects/components/project-settings/components/data-schema/data-schema.component.ts
+++ b/src/app/projects/components/project-settings/components/data-schema/data-schema.component.ts
@@ -1,0 +1,15 @@
+import { Component, OnInit } from '@angular/core';
+
+@Component({
+  selector: 'kern-data-schema',
+  templateUrl: './data-schema.component.html',
+  styleUrls: ['./data-schema.component.scss']
+})
+export class DataSchemaComponent implements OnInit {
+
+  constructor() { }
+
+  ngOnInit(): void {
+  }
+
+}
--- a/src/app/projects/components/project-settings/components/data-schema/data-schema.component.ts
+++ b/src/app/projects/components/project-settings/components/data-schema/data-schema.component.ts
@@ -1,4 +1,14 @@
-import { Component, OnInit } from '@angular/core';
+import { Component, Input, OnInit } from '@angular/core';
+import { FormBuilder, FormGroup } from '@angular/forms';
+import { Observable, Subscription, timer } from 'rxjs';
+import { distinctUntilChanged, first } from 'rxjs/operators';
+import { Project } from 'src/app/base/entities/project';
+import { ProjectApolloService } from 'src/app/base/services/project/project-apollo.service';
+import { WeakSourceApolloService } from 'src/app/base/services/weak-source/weak-source-apollo.service';
+import { attributeVisibilityStates, getTooltipVisibilityState } from '../../../create-new-attribute/attributes-visibility-helper';
+import { DataHandlerHelper } from '../../helper/data-handler-helper';
+import { SettingModals } from '../../helper/modal-helper';
+import { EmbeddingsComponent } from '../embeddings/embeddings.component';
 
 @Component({
   selector: 'kern-data-schema',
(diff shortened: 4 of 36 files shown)


src/app/base/components/dropdown/dropdown.component.ts
```diff
@@ -96,7 +96,6 @@ export class DropdownComponent implements OnChanges {
     if (this.dropdownOptions.isOptionDisabled && this.dropdownOptions.isOptionDisabled.length != this.dropdownOptions.optionArray.length) this.hasInputErrors = "array options != isOptionDisabled length\n";
     if (this.dropdownOptions.optionDescriptions && this.dropdownOptions.optionArray.length != this.dropdownOptions.optionDescriptions.length) this.hasInputErrors = "array options != optionDescriptions length\n";
     if (this.dropdownOptions.optionIcons && this.dropdownOptions.optionIcons.length != this.dropdownOptions.optionIcons.length) this.hasInputErrors = "array options != optionIcons length\n";
-    if (this.dropdownOptions.optionTooltips && this.dropdownOptions.optionTooltips.length != this.dropdownOptions.optionArray.length) this.hasInputErrors = "array options != optionTooltip length\n";
```
JWittmeyer, line 99: why was this removed?

src/app/base/entities/project.ts
```diff
@@ -7,4 +7,6 @@ export interface Project {
   projectType: string;
   description?: string;
   numDataScaleUploaded?: number;
+  tokenizer?: string;
+  containsUniqueAttribute?: boolean;
```
JWittmeyer: ok this one took me a bit :D i didn#t know what this is meant to be, especially since it's never used (only requested for the project by project id request)

since we don't use it in the frontend atm I would just remove it from the query and here. 

In general we should rename the field to something like has_primary_key in the backend (at least that is whats check) -> open an issue on the gateway to rename the field

src/app/model-download/components/model-download-component/model-download-component.component.ts
```diff
@@ -171,9 +169,6 @@ export class ModelDownloadComponentComponent implements OnInit {
     return findModel !== undefined ? true : false;
   }
 
-  goBack() {
-    this.router.navigate(["../" + this.lastPage], { relativeTo: this.activatedRoute });
-  }
```
JWittmeyer, line 176: why was this removed?

src/app/projects/components/project-settings/components/data-schema/data-schema.component.ts
```diff
@@ -0,0 +1,43 @@
...
+
+@Component({
+  selector: 'kern-data-schema',
+  templateUrl: './data-schema.component.html',
+  styleUrls: ['./data-schema.component.scss']
+})
+export class DataSchemaComponent implements OnChanges {
+
+  @Input() project: Project;
+  @Input() attributeVisibilityStates: AttributeVisibilityStates;
```
JWittmeyer: from the Input and name I was expecting an enum with the states. This seems to be the whitelist of possible attribute States for the visibility -> why is this an Input and not just the reference to the const value from attributes visibility helper?

src/app/projects/components/project-settings/components/data-schema/data-schema.component.ts
```diff
@@ -0,0 +1,43 @@
...
+  @Input() pKeyValid: boolean;
+
+  tooltipsArray: string[] = [];
+
+  constructor(private projectApolloService: ProjectApolloService) {
+  }
+
+  ngOnChanges(changes: SimpleChanges) {
+    this.attributes = changes.attributes?.currentValue;
+    this.pKeyValid = changes.pKeyValid?.currentValue;
```
JWittmeyer: You don't need to set the values in the on changes, this will be done automatically. 

Also, this wont work since the changes only hold the values for the ones that did change. So with this approach if pkeyvalid is changed all other variables are set to null since they aren't in the structure. If we for some reason still would need to set them we could do something like:

```
if (changes.attributes) {
    this.attributes = changes.attributes.currentValue
}
```

However as mentioned in the beginning shouldn't be nessecary

src/app/projects/components/project-settings/components/data-schema/data-schema.component.ts
```diff
@@ -0,0 +1,43 @@
...
+    this.attributes.forEach((attribute) => {
+      this.tooltipsArray.push(getTooltipVisibilityState(attribute.visibility));
+    });
```
JWittmeyer: since this.tooltipsArray isn't initialized to empty this will result in an ever-growing array (probably also the reason the warning was disabled in the dropdown component?)

src/app/projects/components/project-settings/components/data-schema/data-schema.component.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  updateVisibility(option: string, attribute: Attribute) {
+    const visibility = attributeVisibilityStates[option].value;
+    this.projectApolloService.updateAttribute(this.project.id, attribute.id, attribute.dataType, attribute.isPrimaryKey, attribute.name, attribute.sourceCode, visibility).pipe(first()).subscribe();
+  }
```
JWittmeyer: old code, no need to provide everything. project-id, attribute id and visibility would do it (since optional parameters in the mutation on backend) should be possible by setting the method variables to null for everything but the nessecary
```suggestion
    this.projectApolloService.updateAttribute(this.project.id, attribute.id, null, null, null, null, visibility).pipe(first()).subscribe();
  }
```

src/app/projects/components/project-settings/components/data-schema/data-schema.component.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  updatePrimaryKey(attribute: Attribute) {
+    this.projectApolloService.updateAttribute(this.project.id, attribute.id, attribute.dataType, !attribute.isPrimaryKey, attribute.name, attribute.sourceCode, attribute.visibility).pipe(first()).subscribe();
```
JWittmeyer: see above

src/app/projects/components/project-settings/components/embeddings/embeddings.component.ts
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  ngOnInit(): void {
+    this.prepareDownloadedModels();
+  }
+
+  ngOnChanges(changes: SimpleChanges) {
+    this.attributesArrayTextUsableUploaded = changes.attributesArrayTextUsableUploaded?.currentValue;
+    this.embeddings = changes.embeddings?.currentValue;
+    this.embeddingHandlesMap = changes.embeddingHandlesMap?.currentValue;
```
JWittmeyer: see comment onchanges

src/app/projects/components/project-settings/components/embeddings/embeddings.component.ts
```diff
@@ -0,0 +1,139 @@
...
+  styleUrls: ['./embeddings.component.scss'],
+})
+export class EmbeddingsComponent implements OnInit {
+
+  @Input() project: Project;
+  @Input() attributesArrayTextUsableUploaded: Attribute[] = [];
+  @Input() settingModals: SettingModals;
+  @Input() isManaged: boolean;
+  @Input() embeddings: Embedding[] = [];
+  @Input() embeddingHandlesMap: Map<string, any> = new Map<string, any>();
```
JWittmeyer: map

JWittmeyer: why are some of these initialzed with empty version? Since they will be overwritten anyway this is overhead. I'm assuming this was done to prevent errors since the values are provided "later" since some requests take some time. However for that we can simply display a loading icon so for onchanges a function call of checkStilLoading or something like that

```
private checkStillLoading(){
this.somethingLoading = !this.attributesArrayTextUsableUploaded || ....
}
```

the flag somethingLoading would then be used in HTML to either display loading or the table body.

src/app/projects/components/project-settings/components/embeddings/embeddings.component.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  downloadedModelsList$: any;
+  downloadedModelsQuery$: any;
+  downloadedModels: DownloadedModel[] = [];
+  subscriptions$: Subscription[] = [];
+  dataHandlerHelper: DataHandlerHelper;
+
+  constructor(private projectApolloService: ProjectApolloService, private formBuilder: FormBuilder) {
+    this.dataHandlerHelper = new DataHandlerHelper(this.formBuilder, this.projectApolloService);
```
JWittmeyer: this is another initialization of the datahandler class. this seems wrong

src/app/projects/components/project-settings/components/embeddings/embeddings.component.ts
```diff
@@ -0,0 +1,139 @@
...
+  @Input() settingModals: SettingModals;
+  @Input() isManaged: boolean;
+  @Input() embeddings: Embedding[] = [];
+  @Input() embeddingHandlesMap: Map<string, any> = new Map<string, any>();
+
+  downloadedModelsList$: any;
+  downloadedModelsQuery$: any;
+  downloadedModels: DownloadedModel[] = [];
+  subscriptions$: Subscription[] = [];
+  dataHandlerHelper: DataHandlerHelper;
```
JWittmeyer: all of these seem a bit fishy:
- downloadedModelsList$: any; -> value changed -> only used once to init subscription local variable would be enough
- downloadedModelsQuery$: any; -> query -> no websocket stuff so never used again
- downloadedModels: DownloadedModel[] = []; initialized with an empty array though reset on every subscription call -> checkIfModelIsDownloaded also checks for undefined so shouldn't be needed to initialze
- subscriptions$: Subscription[] = []; -> are set but never unsubscribed
- dataHandlerHelper: DataHandlerHelper; -> maybe as input if absolutely necessary but not created new (since this will run the constructor again but only the constructor so the necessary variables aren't filled)

src/app/projects/components/project-settings/components/embeddings/embeddings.component.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  ngOnChanges(changes: SimpleChanges) {
+    this.attributesArrayTextUsableUploaded = changes.attributesArrayTextUsableUploaded?.currentValue;
+    this.embeddings = changes.embeddings?.currentValue;
+    this.embeddingHandlesMap = changes.embeddingHandlesMap?.currentValue;
+    this.embeddingHandlesMap.forEach((value: any) => {
+      value.forEach((element: any) => {
+        element.isModelDownloaded = this.checkIfModelIsDownloaded(element.configString);
+      });
+    });
```
JWittmeyer: this only needs to be rechecked if the model or the embedding handles were changed

src/app/projects/components/project-settings/components/embeddings/embeddings.component.ts
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  prepareDownloadedModels() {
+    [this.downloadedModelsQuery$, this.downloadedModelsList$] = this.projectApolloService.getModelProviderInfo();
+    this.subscriptions$.push(
+      this.downloadedModelsList$.subscribe((downloadedModels) => this.downloadedModels = downloadedModels));
```
JWittmeyer: this also needs to recheck the values for embeddingHandlesMap so if a new values is provided by the subscription the values for isModelDownloaded are up to date.

I would create a small function that can be called here as well as in the change method if the ebmedingHandlesMap was changed:

```
checkModelDownloaded(){
    if (!this.embeddingHandelsMap || !this.downloadedModels) return;
    this.embeddingHandlesMap.forEach((value: any) => {
      value.forEach((element: any) => {
        element.isModelDownloaded = this.checkIfModelIsDownloaded(element.configString);
      });
    });

}
```

src/app/projects/components/project-settings/components/embeddings/embeddings.component.html
```diff
@@ -0,0 +1,245 @@
...
+                                </td>
+                            </tr>
+                        </tbody>
+                    </table>
+                </div>
+            </div>
+        </div>
+        <div class="mt-1">
+            <div class="tooltip tooltip-right"
+                data-tip="Vectorize your attributes. Integration to Hugging Face available">
```
JWittmeyer, line 120: tooltip is displayed "below" button -> scss for before element missing imo

src/app/projects/components/project-settings/components/embeddings/embeddings.component.html
```diff
@@ -0,0 +1,245 @@
...
+    buttonCaption: 'Add embedding',
+    disabled: settingModals.embedding.create.blocked || settingModals.embedding.create.embeddingCreationFormGroup?.get('embeddingHandle').value == '',
+    emitObject: this, emitFunction:addEmbedding
+}" (optionClicked)="settingModals.embedding.create.open = false">
+    <div class="flex flex-grow justify-center text-lg leading-6 text-gray-900 font-medium">
+        Add an embedding </div>
+    <div class="mb-2 flex flex-grow justify-center text-sm text-gray-500">
+        Pick from the below solutions to build a vector representation</div>
+    <div *ngIf="attributesArrayTextUsableUploaded && settingModals.embedding.create.embeddingCreationFormGroup">
+        <form [formGroup]="settingModals.embedding.create.embeddingCreationFormGroup"
```
JWittmeyer, line 182: didn't we decide against formgroups? unsure though :D

src/app/projects/components/project-settings/components/embeddings/embeddings.component.ts
```diff
@@ -0,0 +1,139 @@
...
+      if (!embeddingHandle.hidden && !embeddingHandle.forceHidden) {
+        this.selectEmbeddingHandle(embeddingHandle, inputElement);
+        return;
+      }
+    }
+  }
+
+  selectEmbeddingHandle(embeddingHandle: Embedding, inputElement: HTMLInputElement, hoverBox?: any) {
+    inputElement.value = embeddingHandle.configString;
+    hoverBox.style.display = 'none';
```
JWittmeyer: seems to run into erros since hoverbox isn'T defined. Since this is an optional parameter I think we just need to check:
```suggestion
    if(hoverBox) hoverBox.style.display = 'none';
```

JWittmeyer: also an issue in the current version of the app :D

src/app/model-download/components/model-download-component/model-download-component.component.ts
lumburovskalina: we have another method for going back, not sure which approach is better
`  clickBack() {
    RouteManager.moveBack();
  }`

src/app/base/components/dropdown/dropdown.component.ts
lumburovskalina: Returned it back, as you stated in the other comment the tooltipArray was not set to empty.

src/app/projects/components/project-settings/components/embeddings/embeddings.component.ts
```diff
@@ -0,0 +1,155 @@
...
+  styleUrls: ['./embeddings.component.scss'],
+})
+export class EmbeddingsComponent implements OnInit {
+
+  @Input() project: Project;
+  @Input() attributesArrayTextUsableUploaded: Attribute[];
+  @Input() settingModals: SettingModals;
+  @Input() isManaged: boolean;
+  @Input() embeddings: Embedding[];
+  @Input() embeddingHandlesMap: { [key: string]: any };
```
JWittmeyer: Still has Map in the name. Maybe just embeddingHandles or embeddingHandlesDict?

Also the syntax for "any kind of key" is [key:string]:any to represent that the dictionary is a string key and holds something. 

This is not wrong but to better understand the code I recently changed the variable name of key to be a bit more representative -> in this case `[embeddingId:string]:any` so the developer knows that its not "any" key but specifically an embeddingId. 

Again not wrong but might help a dev in a few months to better understand the code :)

src/app/projects/components/project-settings/components/embeddings/embeddings.component.ts
```diff
@@ -0,0 +1,155 @@
...
+
+@Component({
+  selector: 'kern-embeddings',
+  templateUrl: './embeddings.component.html',
+  styleUrls: ['./embeddings.component.scss'],
+})
+export class EmbeddingsComponent implements OnInit {
+
+  @Input() project: Project;
+  @Input() attributesArrayTextUsableUploaded: Attribute[];
```
JWittmeyer: what exactly is this? 
an array of attributes that holds only useable && uploaded && text attributes?

Checking the code this more looks like a `useableTextAttributes` because useable is either useable, uploaded or automatically created.

src/app/projects/components/project-settings/components/embeddings/embeddings.component.ts
```diff
@@ -0,0 +1,155 @@
...
+import { Embedding } from '../../entities/embedding.type';
+import { DataHandlerHelper } from '../../helper/data-handler-helper';
+import { SettingModals } from '../../helper/modal-helper';
+
+@Component({
+  selector: 'kern-embeddings',
+  templateUrl: './embeddings.component.html',
+  styleUrls: ['./embeddings.component.scss'],
+})
+export class EmbeddingsComponent implements OnInit {
```
JWittmeyer: has ondestroy method but not the "interface"

JWittmeyer: same for changes

src/app/projects/components/project-settings/components/data-schema/data-schema.component.ts
```diff
@@ -0,0 +1,61 @@
...
+import { Attribute } from '../../entities/attribute.type';
+import { DataHandlerHelper } from '../../helper/data-handler-helper';
+
+
+@Component({
+  selector: 'kern-data-schema',
+  templateUrl: './data-schema.component.html',
+  styleUrls: ['./data-schema.component.scss']
+})
+export class DataSchemaComponent implements OnInit {
```
JWittmeyer: has changes method but not the interface

src/app/projects/components/project-settings/components/embeddings/embeddings.component.ts
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  prepareDownloadedModels() {
+    let downloadedModelsQuery$, downloadedModelsList$;
+    [downloadedModelsQuery$, downloadedModelsList$] = this.projectApolloService.getModelProviderInfo();
+    this.subscriptions$.push(
+      downloadedModelsList$.subscribe((downloadedModels) => {
+        this.downloadedModels = downloadedModels;
+        this.checkModelDownloaded();
+      }));
```
JWittmeyer: im not sure that this will work as intended. The subscription is created but the query is never refetched e.g. on a websocket notification (`msgParts[1] === 'model_provider_download' && msgParts[2] === 'finished'` )

since this is never refetched it doesn't need to be pushed to notifications. However I think we should use websockets here

src/app/projects/components/project-settings/components/embeddings/embeddings.component.ts
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  checkForceHiddenHandles() {
+    const granularity = this.settingModals.embedding.create.embeddingCreationFormGroup.get('granularity').value;
+    const attId = this.settingModals.embedding.create.embeddingCreationFormGroup.get('targetAttribute').value;
+    const suggestionList = this.embeddingHandlesMap[attId];
+    for (let element of suggestionList) {
+      element = { ...element };
```
JWittmeyer: this copies the element every time the method is called. I think i remember we needed to copy responses from gql if we want to extend the elements (in this case forceHidden) but we don't need to do this every time imo. Doing this once in the subscription (e.g. with `jsoncopy` should do it. I also might be missing something and this is necessary here. just not 100%

src/app/projects/components/project-settings/components/labeling-tasks/labeling-tasks.component.html
```diff
@@ -0,0 +1,408 @@
...
+                                <th scope="col"
+                                    class="px-3 py-2 text-center text-xs font-medium uppercase tracking-wide text-gray-500">
+                                </th>
+                            </tr>
+                        </thead>
+                        <tbody class="divide-y divide-gray-200">
+                            <tr *ngFor="let labelingTaskContainer of labelingTasksArray.controls; let j = index"
+                                [ngClass]="j%2!=0 ? 'bg-gray-50':'bg-white'" [formGroupName]="j">
+                                <td class="whitespace-nowrap text-center px-3 py-2 text-sm text-gray-500">
+                                    {{labelingTaskContainer.get("targetName").value}}
```
JWittmeyer, line 37: because of form group function in HTML

JWittmeyer: can be ignored for now until we refactor form groups in general

src/app/projects/components/project-settings/components/labeling-tasks/labeling-tasks.component.html
```diff
@@ -0,0 +1,408 @@
...
+                                            (keyup)="checkTaskNameColor($event.target)"
+                                            (keydown.enter)="changeTaskName(labelingTaskContainer,inputTaskName.value)"
+                                            (blur)="changeTaskName(labelingTaskContainer,inputTaskName.value)">
+                                    </ng-template>
+                                </td>
+                                <td class="whitespace-nowrap text-center px-3 py-2 text-sm text-gray-500">
+                                    <select formControlName="taskType"
+                                        class="select select-sm select-bordered pr-8 max-w-xs">
+                                        <option *ngFor="let dropdown of labelingTasksDropdownValues()"
+                                            [attr.disabled]="isLabelingTaskOptionDisabled(labelingTaskContainer,dropdown.value)?true:null"
```
JWittmeyer, line 57: function

src/app/projects/components/project-settings/components/labeling-tasks/labeling-tasks.component.html
```diff
@@ -0,0 +1,408 @@
...
+        you want to delete this label?</p>
+    <p class="text-gray-500 text-sm text-center">This will
+        delete
+        all data associated with it!</p>
+</kern-modal>
+
+<kern-modal [isOpen]="settingModals.label.create.open" closeButton="X" [modalBoxStyle]="{'width':'30rem'}"
+    [acceptButton]="{
+    buttonCaption: 'Add label',
+    disabled: !isLabelNameUnique(settingModals.label.create.taskId,labelName.value) || labelName.value == '',
```
JWittmeyer: function in html

src/app/projects/components/project-settings/components/labeling-tasks/labeling-tasks.component.ts
```diff
@@ -0,0 +1,304 @@
...
+import { DataHandlerHelper } from '../../helper/data-handler-helper';
+import { LabelHelper } from '../../helper/label-helper';
+import { SettingModals } from '../../helper/modal-helper';
+
+@Component({
+  selector: 'kern-labeling-tasks',
+  templateUrl: './labeling-tasks.component.html',
+  styleUrls: ['./labeling-tasks.component.scss']
+})
+export class LabelingTasksComponent implements OnInit {
```
JWittmeyer: missing interfaces

src/app/projects/components/project-settings/components/labeling-tasks/labeling-tasks.component.ts
```diff
@@ -0,0 +1,304 @@
...
+    if (this.labelingTasksDropdownArray.length == 0) {
+      for (let t of Object.values(LabelingTask)) {
+        if (t == LabelingTask.NOT_USEABLE) continue;
+        this.labelingTasksDropdownArray.push({
+          name: labelingTaskToString(t),
+          value: t,
+        });
+      }
+    }
+    return this.labelingTasksDropdownArray;
```
JWittmeyer, line 127: this is old logic used to always have an initialized array with the method (common c# style) since we don't want to use function in the HTML anymore it would be better to initialize the labelingTasksDropdownArray in the oninit use it in the HTML and be done with it

src/app/projects/components/project-settings/components/labeling-tasks/labeling-tasks.component.ts
```diff
@@ -0,0 +1,304 @@
...
+  getTaskArrayAttribute(taskId: string, valueID: string) {
+    for (let task of this.labelingTasksArray.controls) {
+      if (taskId == task.get('id').value) return task.get(valueID).value;
+    }
+    return 'UNKNOWN';
+  }
+
+  removeLabel() {
+    const labelDeleteData = this.settingModals.label.delete;
+    this.lh.removeLabel(this.project.id, labelDeleteData.taskId, labelDeleteData.label.id, labelDeleteData.label.color);
```
JWittmeyer, line 245: why does a remove function need a label color? also I think we should at least check for modal open that if it's not open we don't remove a random label

JWittmeyer: found the reason for oldcolor (so the new labels know what colors are already taken)

src/app/projects/components/project-settings/helper/data-handler-helper.ts
```diff
@@ -0,0 +1,154 @@
...
+        const input = document.getElementById(inputBoxName) as HTMLInputElement;
+        if (input && input instanceof HTMLElement) {
+            setTimeout(() => {
+                input.focus();
+            }, 0);
+            return;
+        }
+    }
+
+    getAttributeArrayAttribute(attributeId: string, valueID: string, attributes: Attribute[]) {
```
JWittmeyer: valueID => valueKey

src/app/projects/components/project-settings/helper/data-handler-helper.ts
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    attributeChangedToText(attributes: Attribute[]): boolean {
+        for (let i = 0; i < attributes.length; i++) {
+            const att = attributes[i]
+            const wantedDataType = this.getAttributeArrayAttribute(att.id, 'dataType', attributes);
+            if (att.dataType != wantedDataType && wantedDataType == "TEXT") return true;
+        }
+        return false;
+    }
```
JWittmeyer: i think this cant be done anymore so we don't need the function

src/app/projects/components/project-settings/helper/data-handler-helper.ts
```diff
@@ -0,0 +1,154 @@
...
+
+    createAttributeTokenStatistics(projectId: string, attributeId: string) {
+        this.projectApolloService.createAttributeTokenStatistics(projectId, attributeId).pipe(first()).subscribe();
+    }
```
JWittmeyer: see above

src/app/projects/components/project-settings/helper/label-helper.ts
```diff
@@ -75,8 +75,11 @@ export class LabelHelper {
 
         labelInput.value = '';
         labelInput.focus();
-        this.settings.requestTimeOut = true;
-        timer(100).subscribe(() => this.settings.requestTimeOut = false);
+        requestTimeOut = true;
```
JWittmeyer: requestTimout is called by value so changes aren't set "sent back" to the calling function. Either use a wrapper object or return logic

src/app/projects/components/project-settings/helper/label-helper.ts
```diff
@@ -75,8 +75,11 @@ export class LabelHelper {
...
+        timer(100).subscribe(() => {
+            requestTimeOut = false;
+            return requestTimeOut;
```
JWittmeyer: this is not a return for the add label function but for the provided anonymous function in {}

src/app/projects/components/project-settings/helper/label-helper.ts
JWittmeyer: i think only by refernce can work because of the timer stuff later

JWittmeyer:
renaming a variable doesnt change it in the view directly only after a refresh

- [ ] resolved
- [x] didn't occur on my side

src/app/model-download/components/model-download-component/model-download-component.component.ts
lumburovskalina: Leaving the current clickBack() method and resoling the conversation

src/app/projects/components/project-settings/components/embeddings/embeddings.component.html
lumburovskalina: Resolving the conversation -> removing form groups should be done in the next refactoring steps

src/app/projects/components/project-settings/components/embeddings/embeddings.component.ts
lumburovskalina: Yes, exactly that and it makes sense to rename it

src/app/projects/components/project-settings/components/embeddings/embeddings.component.ts
lumburovskalina: This was added because of some changes for the compilation time, but now I removed it and it still works as expected :D

JWittmeyer:
Project title exist check isn't checked live anymore
image

- [ ] resovled

JWittmeyer:
tokenization config strings are set multiple times?
image

-> create project -> go back -> create project -> open dropdown

- [ ] resolved

JWittmeyer:
Dropdown caption after selection of a value is only the config string not the name + config as before

image


- [ ] resolved

src/app/projects/components/project-settings/components/labeling-tasks/labeling-tasks.component.html
lumburovskalina: I am resolving the conversation because it will be solved by other issue and the function in the html will not exist

src/app/projects/components/project-settings/helper/label-helper.ts
```diff
@@ -75,8 +75,14 @@ export class LabelHelper {
 
         labelInput.value = '';
         labelInput.focus();
-        this.settings.requestTimeOut = true;
-        timer(100).subscribe(() => this.settings.requestTimeOut = false);
+
+        let subject = new Subject<boolean>();
```
JWittmeyer: is there a specific reason why this is now with a subscription solved? using an object reference as stated before should do the same without the need to handle the subscriptions in any specific way. Also the subscription seems to be open indefinitely though it is only used once.

JWittmeyer:
Embedding progress isn't updated anymore.

Best guess: Reassignment of e doesn't make sense (`e = {...e}`). Reassigning the variable "space" creates a new object but it isn't placed in the array => the array doesn't change => the changes aren't transferred to the subcomponent.

With the current collection of embeddings the values aren't changeable (I assume that's the reason why the data was copied like mentioned above) => a jsoncopy should do it:

`this.embeddings = JSON.parse(JSON.stringify((res[1])));`

-->tested and works as expected (didn't check the changes in though :) )

- [x] resolved

src/app/projects/components/project-settings/project-settings.component.ts
```diff
@@ -1,169 +1,138 @@
...
+      this.project = project;
+      [this.attributesQuery$, this.attributes$] = this.dataHandlerHelper.prepareAttributesRequest(projectId);
+      [this.embeddingQuery$, this.embeddings$] = this.dataHandlerHelper.prepareEmbeddingsRequest(projectId);
+      this.suggestions$ = this.projectApolloService.getRecommendedEncodersForEmbeddings(projectId);
+      let tasks$ = [];
+      tasks$.push(this.attributes$);
+      tasks$.push(this.embeddings$);
+      tasks$.push(this.suggestions$);
+
+      combineLatest(tasks$).subscribe((res: any[]) => {
```
JWittmeyer, line 99: why is this done inside the project subscription? so whenever the project is subscription is setup the attribute & embedding querys are resetup. If we necessarily need the project for the rest we could work with a pipe first but it doesn't seem to be necessary

JWittmeyer (approved):

PR #100, opened 2022-12-13
labeling suite

JWittmeyer:
- https://github.com/code-kern-ai/refinery-ui/pull/100
- https://github.com/code-kern-ai/refinery-gateway/pull/99


jhoetter:
image

I don't really understand this bar, i.e. just from looking at the buttons, not sure what they do.

jhoetter:
image

Not 100% sure yet how to really go into details, but I think we really need to figure out a smooth UX when it comes to using these three segments of the new view (i.e. the label legend, the record itself, and the metadata table).

For instance, I could imagine that the legend is collapsable, the record takes up 70% of the view, and the metadata table the rest - and this ratio can be changed/swapped or something like this. Then they should be scrollable, and hovering in the metadata table automatically scrolls to the part in the record where i can see the linked entity.

jhoetter:
image

I like the hovering part a lot, but it isn't clear immediately hovering which part of the row results in which highlighting. if i understood it correctly, hovering the glass icon highlights the one label span, hovering the label name itself also highlights the other spans of that label, and hovering anything else highlights all labels?

Makes sense to me to do something like that, but wasn't intuitive from the table itself.

jhoetter:
image
clicking on the legend to show the context menu wasn't really intuitive for the reason that my mind always wants to trick me that clicking on this legend would result in some kind of label action (as the button as exactly the same). Again, not sure yet how exactly to solve this, but rather trying to just narrow down some thoughts :)

jhoetter:
image
This makes lots of sense to me. Feels like an approach like this could solve two of the above-addressed issues (i.e. not knowing what the buttons are about, and having to find them in the first place). Of course, this is a generalistic bar which would then address all labels at once, but I think that in most cases this should do the trick.



though I'm not fully aligned with how the buttons are built. For instance, not sure why yes/no have each their own button (i.e. why are there two "manual" buttons, if one button indicating a toggle could do the trick?)

jhoetter:
image

I must say I really like these patterns in the background. Would argue which ones to choose, but the concept makes sense to me. Have to dig into my data vis lecture from two years ago to make an educated bet on which combination of border-styling and patterns would work best intuitively, haha :D but there are actually some behavioral patterns we could make use of so this feels really natural.

jhoetter:
image

big fan of the settings modal. Though I'm not sure if the naming is perfectly intuitive. E.g. "quick buttons" got me thinking for some time, as "quick buttons" doesn't describe _what_ the buttons do... just that they are quick :)

should we maybe then hide the other settings (e.g. hide heuristic labels from overview table and such also in the settings modal?)

jhoetter:
image

I'm not sure if this "agressive"-coloring looks good, as it feels like using an accent color for filling everything. On the other hand, shows quite well what parts are actually activated. So, need to think about this :)

jhoetter:
image
nice solution, I love that approach. if the number of labeling tasks grows too much, I think we need to make the Task as a tab itself (e.g. you need to move right/left with some chevron icons to move to another labeling task). but this is only the case if you have way too many labeling tasks and labels. Just something to keep in mind.

jhoetter:
image
really happy how this turned out. I like it.

jhoetter:
image
just as i can hide the label legend, I would suggest to enable hiding the table. For small records, I really wouldn't need it (personally).

jhoetter:
> image
> 
> just as i can hide the label legend, I would suggest to enable hiding the table. For small records, I really wouldn't need it (personally).



stupid me. already implemented :D

image

jhoetter:
For the above, however indicates again that "show feature" isn't a good naming.

jhoetter:
after playing around with it a little bit, I can safely tell:

- it has improved a ton. Like really, that is so much better already than before!
- there definitely is some learning curve. It wasn't all intuitive to me at first, but after playing around with it, I got used to it rather quickly. but i was also meant to test this and play around with it, which isn't the case for a normal user
- i think that fixing some of the naming would help a lot, as well as using toggles instead of buttons for the display of span categories, and improving the hovering in that it becomes more intuitive. we can look into some ideas together.



Awesome job @JWittmeyer

src/app/labeling-suite/helper/manager/record.ts
```diff
@@ -0,0 +1,266 @@
...
+                this.setDeletedState();
+            }
+        } else if (msgParts[1] == 'rla_created' || msgParts[1] == 'rla_deleted') {
+            if (this.recordRequests.rlaQuery) this.recordRequests.rlaQuery.refetch();
+        }
+        else {
+            console.log("unknown message in labeling suite record manager" + msgParts);
+        }
+
+    }
```
FelixKirsch: `payload_finished` and `weak_supervision_finished` are part of the websocket whitelist, but are not in the handle notification method. -> leads to console log

src/app/labeling-suite/helper/manager/recordRla.ts
```diff
@@ -0,0 +1,203 @@
...
+            if ((rla.sourceId && rla.sourceId == currentSourceId)) {
+                rla.sourceType = LabelSource.MANUAL;
+                rla.sourceId = null;
+            } else if (this.baseManager.userManager.roleAssumed && rla.createdBy == userId && rla.labelingTaskLabel.labelingTask.id == allowedTask
+                && rla.sourceType == LabelSource.MANUAL) {
+
+            } else {
+                rla.id = "x";
+            }
+        });
```
FelixKirsch: The `else if` contains no code. Is this intentional to define which case should be run in the `else` clause? Then I would prefer negating the `else if` condition and adding the `else` code there.

src/app/labeling-suite/helper/manager/recordRla.ts
```diff
@@ -0,0 +1,203 @@
...
+        return rlaData.filter(rla => rla.id != "x");
+    }
+
+    private getTokenizedAttribute(attributeId: string, tokenAttribute: any) {
```
FelixKirsch: typing, return value

src/app/labeling-suite/helper/manager/recordRla.ts
```diff
@@ -0,0 +1,203 @@
...
+        for (let att of tokenAttribute) {
+            if (att.attributeId == attributeId) return att;
+        }
+        return null;
+    }
+
+    private getToken(tokenizedAttribute, idx: number) {
```
FelixKirsch: typing, return value

src/app/labeling-suite/helper/manager/recordRla.ts
```diff
@@ -0,0 +1,203 @@
...
+    private getCreatedByName(e: any): string {
+        if (e.sourceType == LabelSource.INFORMATION_SOURCE) return e.informationSource.name;
+        if (!e.createdBy || e.createdBy == "NULL") return '-';
+        else if (!e.user?.firstName) return 'Unknown User ID';
+        else {
+            return e.user.firstName + ' ' + e.user.lastName;
+        }
+    }
+
+    private getLabelData(e: any) {
```
FelixKirsch: typing, return value

src/app/labeling-suite/helper/manager/recordRla.ts
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    private getLabelDataTip(e: any): string {
+        if (e.sourceType == LabelSource.INFORMATION_SOURCE) return informationSourceTypeToString(e.informationSource.type, false);
+        else if (e.sourceType == LabelSource.WEAK_SUPERVISION) return "Weak supervision - click to use as manual label";
+        else if (e.sourceType == LabelSource.MANUAL) return "Manual";
+        return null;
+    }
+
+    private getLabelForDisplay(e: any) {
```
FelixKirsch: typing, return value

src/app/labeling-suite/helper/manager/recordRla.ts
```diff
@@ -0,0 +1,203 @@
...
+        let final = e.labelingTaskLabel.name;
+        if (e.sourceType == LabelSource.WEAK_SUPERVISION && e.confidence != null) {
+            final += " - " + Math.round((e.confidence + Number.EPSILON) * 10000) / 100 + '%';
+        } else if (this.baseManager.settingManager.settings.labeling.showHeuristicConfidence && e.sourceType == LabelSource.INFORMATION_SOURCE && e.confidence != null) {
+            final += " - " + Math.round((e.confidence + Number.EPSILON) * 10000) / 100 + '%';
+        }
+        return final;
```
FelixKirsch, line 184: In both cases, `if` an `else if`, the executed code is the same.

src/app/labeling-suite/helper/manager/record.ts
```diff
@@ -0,0 +1,266 @@
...
+        this.baseManager.runUpdateListeners(UpdateType.RECORD);
+    }
+
+    doBeforeDestroy(): void {
+        for (const key in this.recordRequests) {
+            if (key == 'rlaQuery') continue;
+            if (this.recordRequests[key]) this.recordRequests[key].unsubscribe();
+        }
+        NotificationService.unsubscribeFromNotification(this, this.projectId);
+    }
```
FelixKirsch, line 65: How is our approach to typing in typescript if a function is not returning anything?
Do we want to add `void` everywhere?

src/app/labeling-suite/helper/manager/session.ts
```diff
@@ -0,0 +1,394 @@
...
+
+export enum LabelingLinkType {
+    SESSION = 'SESSION',
+    DATA_SLICE = 'DATA_SLICE',
+    HEURISTIC = 'HEURISTIC',
+}
+const ONE_DAY = 86400000; // 24 * 60 * 60 * 1000;
+const DUMMY_HUDDLE_ID = "00000000-0000-0000-0000-000000000000";
+
+export function parseLabelingLinkData(route: ActivatedRoute): LabelingLinkData {
```
FelixKirsch, line 42: typing, return value

src/app/labeling-suite/helper/manager/session.ts
```diff
@@ -0,0 +1,394 @@
...
+        if (['access_link_changed', 'access_link_removed'].includes(msgParts[1])) {
+            if (this.router.url.indexOf(msgParts[3]) > -1 && this.labelingLinkData) {
+                //python "True" string
+                this.labelingLinkData.linkLocked = !msgParts[4] || msgParts[4] === 'True';
+                location.reload();
+            }
+        }
+    }
+
+    public getAllowedTaskId() {
```
FelixKirsch: typing, return value

src/app/labeling-suite/helper/manager/record.ts
JWittmeyer: good catch -> included in rla query refetch

src/app/labeling-suite/helper/manager/session.ts
```diff
@@ -0,0 +1,394 @@
...
+        if (this.debounceTimer) this.debounceTimer.unsubscribe();
+        const jumpIdx = jumpPos - 1;
+        this.debounceTimer = timer(200).subscribe(() => this.baseManager.recordManager.collectRecordData(this.huddleData.recordIds[jumpIdx]));
+        this.nextDisabled = !this.huddleData || jumpPos == this.huddleData.recordIds.length;
+        this.prevDisabled = !this.huddleData || jumpIdx == 0;
+        this.positionString = jumpPos + " / " + this.huddleData.recordIds.length + " records in";
+    }
+
+
+    private parseCheckedAt(checkedAt: string) {
```
FelixKirsch: typing, return value

src/app/labeling-suite/helper/manager/session.ts
```diff
@@ -0,0 +1,394 @@
...
+                if (!this.stillInLabeling()) return;
+                this.availableLinks = availableLinks;
+                this.availableLinksLookup = {};
+                this.availableLinks.forEach(link => this.availableLinksLookup[link.id] = link);
+                const linkRoute = this.router.url.split("?")[0];
+                this.selectedLink = this.availableLinks.find(link => link.link.split("?")[0] == linkRoute);
+            });
+    }
+
+    private stillInLabeling() {
```
FelixKirsch: typing, return value

src/app/labeling-suite/helper/manager/session.ts
```diff
@@ -0,0 +1,394 @@
...
+        if (!RouteManager.currentUrl) return true;
+        return RouteManager.currentUrl.includes("labeling");
+    }
+
+    private checkLocalDataOutdated() {
```
FelixKirsch: typing, return value

src/app/labeling-suite/helper/manager/session.ts
```diff
@@ -0,0 +1,394 @@
...
+        pipeFirst.subscribe((outdated) => {
+            if (outdated) {
+                localStorage.removeItem("huddleData");
+                this.huddleData = null;
+            }
+        });
+        return pipeFirst;
+    }
+
+    private checkLinkAccess() {
```
FelixKirsch: typing, return value

src/app/labeling-suite/main-component/labeling-suite.component.scss
```diff
@@ -0,0 +1,87 @@
...
+  pointer-events: none;
+  border-radius: 0.375rem;
+  opacity: 0.3;
+}
+
+
+.wait,
+.wait * {
+  cursor: wait;
+}
```
FelixKirsch: already defined on top of file

src/app/labeling-suite/sub-components/labeling/helper.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+type TaskLookup = {
+    showText: boolean;
+    showGridLabelPart: boolean;
+    goldInfo?: {
+        can: boolean,
+        is: boolean,
+    };
+    girdRowSpan?: string;
```
FelixKirsch: typo `girdRowSpan`

src/app/labeling-suite/sub-components/labeling/labeling.component.ts
```diff
@@ -0,0 +1,674 @@
...
+  private rlaIsEqual(rlaA: any, rlaB: any): boolean {
+    if (rlaA.sourceType != rlaB.sourceType) return false;
+    if (rlaA.returnType != rlaB.returnType) return false;
+    if (rlaA.labelingTaskLabelId != rlaB.labelingTaskLabelId) return false;
+    if (rlaA.returnType == InformationSourceReturnType.YIELD && (rlaA.tokenStartIdx != rlaB.tokenStartIdx || rlaA.tokenEndIdx != rlaB.tokenEndIdx)) return false;
+
+    return true;
+  }
+
+  private getTokenData(attributeId: string) {
```
FelixKirsch: typing, return value

src/app/labeling-suite/sub-components/overview-table/overview-table.component.ts
```diff
@@ -0,0 +1,116 @@
...
+        this.headerHover.taskCollection += data.hoverGroups.task.split(',')[0] + ', ';
+        this.headerHover.createdByCollection += data.hoverGroups.createdBy.split(',')[0] + ', ';
+        this.headerHover.labelCollection += data.hoverGroups.label.split(',')[0] + ', ';
+        this.headerHover.rlaCollection += data.hoverGroups.rlaId.split(',')[0] + ', ';
+      }
+    }
+    // const newClass = this.lsm.settingManager.settings.main.hoverGroupBackgroundColorClass + 'border-l border-r bg-gray-200 font-bold';
+    // if (this.headerHover.class !== newClass) {
+    //   this.headerHover.class = newClass;
+    // }
```
FelixKirsch: commented code. Is this still needed or just remained from development?

src/app/util/helper-functions.ts
```diff
@@ -211,4 +212,43 @@ export function getPythonClassName(codeToCheck: string): string {
 
 export function getPythonClassRegExMatch(codeToCheck: string): any {
     return /class ([\w]+)\([^)]+\):/.exec(codeToCheck);
+}
+
+export function jsonCopy(src) {
```
FelixKirsch: typing, return value

FelixKirsch (approved):
looks good :)

src/app/labeling-suite/helper/manager/recordRla.ts
JWittmeyer: is intential -> however I prefer this version since the else case is imo easier to read like this "remove all others" is the final idea => ill add a comment to reduce confusion though :)

src/app/labeling-suite/helper/manager/recordRla.ts
JWittmeyer: changed

src/app/labeling-suite/helper/manager/recordRla.ts
JWittmeyer: changed

src/app/labeling-suite/helper/manager/recordRla.ts
JWittmeyer: changed

src/app/labeling-suite/helper/manager/recordRla.ts
JWittmeyer: changed

src/app/labeling-suite/helper/manager/recordRla.ts
JWittmeyer: true however for readability reasons i splitted it since the condition was unreadable otherwise => I can change it though

src/app/labeling-suite/helper/manager/record.ts
JWittmeyer: the ones from interfaces (implements xxxx) are usually with a void autocompleted, the rest is without return type so ill keep it (and ensure the other interface thing get the void where it's missing)

src/app/labeling-suite/helper/manager/recordRla.ts
FelixKirsch: no that's fine. Just wanted to make sure that it is as intended.

src/app/labeling-suite/sub-components/overview-table/overview-table.component.ts
JWittmeyer: removed

PR #102, opened 2023-01-04
Refactor upload component

lumburovskalina:
Implement: 
- https://github.com/code-kern-ai/refinery/issues/193
- https://github.com/code-kern-ai/refinery/issues/197


src/app/import/components/helpers/upload-helper.ts
```diff
@@ -0,0 +1,72 @@
...
+import { UploadComponent } from "../upload/upload.component";
+
+export class UploadHelper {
+    baseComponent: UploadComponent;
+    recordNewUploadHelper: RecordNewUploadHelper;
+    recordAddUploadHelper: RecordAddUploadHelper;
+    existingProjectUploadHelper: ExistingProjectUploadHelper;
+    lookupListsUploadHelper: LookupListsUploadHelper;
+
+    constructor(baseComponent?: UploadComponent, recordNewUploadHelper?: RecordNewUploadHelper, recordAddUploadHelper?: RecordAddUploadHelper, existingProjectUploadHelper?: ExistingProjectUploadHelper, lookupListsUploadHelper?: LookupListsUploadHelper) {
```
JWittmeyer: im not sure i understand this correctly but the constructor of the UploadHelper wants e.g. an RecordAddUploadHelper. The RecordAddUplaodHelper has in the constructor an uploadHelper variable that is created with new.

Doesn't seem right to me

src/app/import/components/helpers/upload-helper.ts
```diff
@@ -0,0 +1,72 @@
...
+            case UploadFileType.PROJECT:
+                this.existingProjectUploadHelper.doUpload();
+                break;
+            case UploadFileType.KNOWLEDGE_BASE:
+                this.lookupListsUploadHelper.doUpload();
+                break;
+        }
+    }
+
+    getFinalFileName(projectId: string, uploadFileType: UploadFileType, knowledgeBaseId?: string): string {
```
JWittmeyer: the get method sets project id, an execute & a resubscirption? if these are necessary steps the calling methods should do that but not the file name getter => the next dev who tries to get the file name will be irritated

src/app/import/components/helpers/upload-helper.ts
```diff
@@ -0,0 +1,72 @@
...
+                return fileName + "_" + knowledgeBaseId;
+            default:
+                return fileName;
+        }
+    }
+
+    setProjectId(projectId: string, baseComponent: UploadComponent): void {
+        this.baseComponent = baseComponent;
+        this.baseComponent.projectId = projectId;
+    }
```
JWittmeyer: sets the projectid and the base component which was already provided in the constructor. If we need a setter for it we should introduce a specific method for it

src/app/import/components/modal-upload/modal-upload.component.ts
```diff
@@ -0,0 +1,103 @@
...
+  title: string;
+  subTitle: string;
+
+  constructor() { }
+
+  ngOnInit(): void {
+    NotificationService.subscribeToNotification(this, {
+      projectId: this.projectId,
+      whitelist: ['file_upload'],
+      func: this.handleWebsocketNotification
```
JWittmeyer: needs to be unsubscibed on destroy

src/app/import/components/modal-upload/modal-upload.component.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  ngAfterViewInit(): void {
+    this.uploadHelper = new UploadHelper(this.baseComponent, this.baseComponent.recordNewUploadHelper, this.baseComponent.recordAddUploadHelper, this.baseComponent.existingProjectUploadHelper, this.baseComponent.lookupListsUploadHelper);
+  }
+
+  ngOnChanges(changes: SimpleChanges): void {
+    if (changes.isModalOpen) {
+      this.uploadModals.uploadFile.open = changes.isModalOpen.currentValue;
+      this.title = getTitle(this.uploadFileType);
+      this.subTitle = getSubtitle(this.uploadFileType);
```
JWittmeyer: why are these reset if the open state changes?

src/app/import/components/modal-upload/modal-upload.component.ts
```diff
@@ -0,0 +1,103 @@
...
+    }
+  }
+
+  closeModal(): void {
+    this.uploadModals.uploadFile.open = false;
+    this.closeModalEvent.emit();
+    if (this.uploadOptions.knowledgeBaseId) {
+      this.refetchLookupLists.emit(true);
+    }
```
JWittmeyer: shouldn't this be something the dev can decide on their own in the closeModalEvent?

src/app/import/components/modal-upload/modal-upload.component.ts
```diff
@@ -0,0 +1,103 @@
...
+    this.uploadHelper.upload(this.file);
+  }
+
+  setFile(file: File): void {
+    this.file = file;
+  }
+
+  optionClicked(button: string): void {
+    if (button == 'CLOSE') this.closeModal();
+    if (button == 'ACCEPT') {
```
JWittmeyer: else if since it can never be both

src/app/import/components/modal-upload/modal-upload.component.ts
```diff
@@ -0,0 +1,103 @@
...
+        }
+      });
+    }
+  }
+
+  handleWebsocketNotification(msgParts: string[]): void {
+    if (msgParts[1] === 'file_upload' && msgParts[4] === UploadStates.ERROR) {
+      this.baseComponent.deleteExistingProject();
+    }
+  }
```
JWittmeyer: why does the modal component handle websockets? shouldn't this be something the base component does?

src/app/import/components/upload/upload.component.ts
```diff
@@ -1,184 +1,162 @@
...
-  @Input() projectId: string;
-  @Input() reloadOnFinish: boolean = true;
-  @Input() unique;
-  @Input() init = false;
-  @Input() deleteProjectOnFail = false;
-  file: File | null = null;
+export class UploadComponent implements OnInit {
+
+  @Input() uploadFileType: UploadFileType;
+  @Input() projectId?: string;
```
JWittmeyer, line 24: why the questionmark?

src/app/import/components/upload/upload.component.ts
```diff
@@ -1,184 +1,162 @@
...
+  submitted: boolean = false;
+  disableInput: boolean = false;
+  tokenizerValuesDisabled: boolean[] = [];
+
+  constructor(private projectApolloService: ProjectApolloService, private router: Router, private s3Service: S3Service) {
+    this.recordNewUploadHelper = new RecordNewUploadHelper(this.projectApolloService, this);
+    this.recordAddUploadHelper = new RecordAddUploadHelper(this);
+    this.existingProjectUploadHelper = new ExistingProjectUploadHelper(this.projectApolloService, this);
+    this.lookupListsUploadHelper = new LookupListsUploadHelper(this)
+    this.uploadHelper = new UploadHelper(this, this.recordNewUploadHelper, this.recordAddUploadHelper, this.existingProjectUploadHelper, this.lookupListsUploadHelper);
```
JWittmeyer: as mentioned above, this procedure seems wrong -> the uploadHelper class is created 5 times in these lines.

Also why do we need to create all of them even if only one is correct for the provided upload type?

src/app/import/components/upload/upload.component.ts
```diff
@@ -1,184 +1,162 @@
...
+    if (this.uploadType == UploadType.LABEL_STUDIO) {
+      if (this.labelStudioUploadAssistant?.states.preparation != PreparationStep.MAPPING_TRANSFERRED) {
+        this.deleteExistingProject();
+      }
     }
   }
 
-  ngAfterViewInit(): void {
-    this.noFileUploaded.nativeElement.classList.add('block');
+  ngOnChanges(changes: SimpleChanges): void {
```
JWittmeyer, line 68: interface missing

src/app/import/components/upload/upload.component.ts
```diff
@@ -1,184 +1,162 @@
...
-  reSubscribeToNotifications() {
-
-    NotificationService.unsubscribeFromNotification(this, null);
-    NotificationService.subscribeToNotification(this, {
-      projectId: this.projectId,
-      whitelist: ['file_upload'],
-      func: this.handleWebsocketNotification
-    });
-
+  ngOnDestroy(): void {
```
JWittmeyer, line 84: interface missing

src/app/import/components/upload/upload.component.ts
```diff
@@ -1,184 +1,162 @@
...
-      .pipe(first()).subscribe();
-    this.projectApolloService
-      .updateProjectStatus(
-        this.projectId,
-        ProjectStatus.INIT_COMPLETE
-      ).pipe(first()).subscribe();
+  createProject(): Observable<any> {
+    return this.projectApolloService
+      .createProject("Imported Project", "Created during file upload " + this.file.name).pipe(first());
   }
```
JWittmeyer: what is this used for?

src/app/import/components/upload/upload.component.ts
```diff
@@ -1,184 +1,162 @@
...
 
-  createEmptyProject() {
+  updateTokenizerAndProjectStatus(projectId: string): void {
+    this.projectApolloService.changeProjectTokenizer(projectId, this.recordNewUploadHelper.selectedTokenizer).pipe(first()).subscribe();
+    this.projectApolloService.updateProjectStatus(projectId, ProjectStatus.INIT_COMPLETE).pipe(first()).subscribe();
+  }
```
JWittmeyer: i would expect this to be in the upload helper class -> its also only called from there so why not make it a private function in the class?

JWittmeyer:
I think we can reduce the distance between the subtitle and the upload component (e.g. remove the margin bottom for the subtitle):
image


- [x] resolved

JWittmeyer:
Project title exist check isn't checked live anymore
image

- [x] resovled

JWittmeyer:
tokenization config strings are set multiple times?
image

-> create project -> go back -> create project -> open dropdown

- [x] resolved

JWittmeyer:
Dropdown caption after selection of a value is only the config string not the name + config as before

image


- [x] resolved

JWittmeyer:
submitting without a name -> enter name -> error stays (probably fixed with check on enter from previous note)
image

- [x] resolved

JWittmeyer:
upload new page seems to stay in submitted state after an error file was uploaded
image
[dummy-error.zip](https://github.com/code-kern-ai/refinery-ui/files/10457618/dummy-error.zip)

file required message should be cleared with the file imo

- [x] resolved

JWittmeyer:
during upload (blue progress) the file required message is visible aswell
image

- [x] resolved

src/app/import/components/upload/upload.component.ts
```diff
@@ -1,184 +1,162 @@
...
-              this.uploadStarted = true;
-              this.startProgressCall(t.id);
-              return;
-            }
-          }
-        });
+    if (this.uploadType == UploadType.LABEL_STUDIO) {
+      if (this.labelStudioUploadAssistant?.states.preparation != PreparationStep.MAPPING_TRANSFERRED) {
+        this.deleteExistingProject();
+      }
```
JWittmeyer: this doesn't make sense in the OnInit. The idea is that if the component is destroyed but the label studio import already created a "dummy project" it should be deleted

-> basically project new >label studio import > upload a file but don't proceed further (project will be created) > leave the page (destroy should be called after implementation of the interface) > project husk should be deleted 

it worked at some point but in dev it also doesn't work as expected :D 

if there are questions don't hesitate :)

JWittmeyer:
error in label studio import resutls in unclickable button 

image

The inputs are all disabled after the initial file was uploaded with the label studio import > this is correct. The button however suggest there is action possible. on dev "nothing" is show
image

also the file required message is wrong (but I'm guessing that it's related to a before-mentioned error)

- [x] resolved

JWittmeyer:
snapshot upload closes the modal before its finished (only green progress is shown)

- [x] resolved

src/app/import/components/modal-upload/modal-upload.component.ts
lumburovskalina: Moved them to on init

src/app/import/components/modal-upload/modal-upload.component.ts
lumburovskalina: Yes, I am adding new option to the uploadOptions as closeModalOnCick

src/app/import/components/modal-upload/modal-upload.component.ts
```diff
@@ -0,0 +1,87 @@
...
+export class ModalUploadComponent implements OnInit, OnChanges {
+
+  @Input() uploadFileType: UploadFileType;
+  @Input() isModalOpen: boolean;
+  @Input() uploadOptions: UploadOptions;
+  @Input() projectId: string;
+
+  @Output() closeModalEvent = new EventEmitter();
+  @Output() fileAttached = new EventEmitter<File>();
+  @Output() refetchLookupLists = new EventEmitter<boolean>();
```
JWittmeyer: this is still an output event which I don't understand => shouldn't exist, the component using the modal-uploadl component should decide when it wants to refetch (on normal close event)

JWittmeyer:
it's a bit irritating for me to see 0 B after I uploaded something with a definite file size 
image

can we store the size and display it after upload or remove the size so the 0 B aren't shown after the upload was initiated?

- [x] resolved

JWittmeyer (approved):

PR #103, opened 2023-01-05
Record-data shared component

lumburovskalina:
Implements:
- https://github.com/code-kern-ai/refinery/issues/196


src/app/base/components/record-card/record-card.component.ts
```diff
@@ -0,0 +1,70 @@
...
+  selector: 'kern-record-card',
+  templateUrl: './record-card.component.html',
+  styleUrls: ['./record-card.component.scss']
+})
+export class RecordCardComponent implements OnInit {
+
+  @Input() record: Record;
+  @Input() attributesSortOrder: AttributeSort;
+  @Input() index: number;
+  @Input() extendedRecords: ExtendedRecord;
```
JWittmeyer: this should be singular, right? we only display one record per card

src/app/base/components/record-card/record-card.component.ts
```diff
@@ -0,0 +1,70 @@
...
+    const saveSimilaritySearch = this.dataBrowserModals.similaritySearch;
+    this.similarSearchHelper.requestSimilarSearch(saveSimilaritySearch.embeddingId, saveSimilaritySearch.recordId);
+  }
+
+  setEmbeddingIdSS(selectedValue: string) {
+    this.dataBrowserModals.similaritySearch.embeddingId = this.similarSearchHelper.embeddings[selectedValue].id;
+  }
+}
+
+export type Record = {
```
JWittmeyer: I'd put the types into their own .ts file. Also, Record feels a bit too general imo. Since this is specifically needed for the databrowser and only works in the databrowser (or at least in combination with the extended record) I'd go with either ExtendedRecord or DatabrowserRecord as type name

src/app/base/components/record-card/record-card.component.ts
JWittmeyer: ok i might have misread something here. Why is the extendedRecord needed? (the structure to hold all records in the record specific component)

src/app/base/components/record-card/record-card.component.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  @Output() preliminaryRecordIds = new EventEmitter<number>();
+
+  constructor() { }
+
+  ngOnInit(): void {
+  }
+
+  storePreliminaryRecordIds(index: number) {
+    this.preliminaryRecordIds.emit(index);
```
JWittmeyer: why is that emitted? if this only stores it in the local storage so maybe no need to emit?

src/app/base/components/record-card/record-card.component.ts
```diff
@@ -0,0 +1,70 @@
...
+  @Input() attributes: Map<string, any>;
+  @Input() dataBrowserModals: DataBrowserModals;
+  @Input() activeSearchParams: any;
+
+  @Output() preliminaryRecordIds = new EventEmitter<number>();
+
+  constructor() { }
+
+  ngOnInit(): void {
+  }
```
JWittmeyer: needed?

src/app/base/components/record-card/record-card.component.ts
JWittmeyer: I see that the session ID is used somewhere in the HTML -> we can of course provide that. 


So in general a bit unclear on the direction we want to go with this component.
This could be A: To reuse the component (e.g. in the display for the "View" for run on 10 for heuristics) then we need to generalize a bit more (like record & attribute arrays and some optional stuff) 
Or B: Use this only as a code restructure tool only meant for the databrowser -> then we should organize it below the data browser in the structure of the folders


-> I think I would go with A to further reduce the complexity where we can -> but we can talk about this

src/app/base/components/record-card/record-card.component.ts
```diff
@@ -0,0 +1,70 @@
...
+  @Input() attributesSortOrder: AttributeSort;
+  @Input() index: number;
+  @Input() extendedRecords: ExtendedRecord;
+  @Input() similarSearchHelper: any;
+  @Input() recordComments: RecordCommentsModal;
+  @Input() attributes: Map<string, any>;
+  @Input() dataBrowserModals: DataBrowserModals;
+  @Input() activeSearchParams: any;
+
+  @Output() preliminaryRecordIds = new EventEmitter<number>();
```
JWittmeyer: the name suggests an array, the emitter is for a single number

src/app/base/components/record-card/record-card.component.ts
```diff
@@ -0,0 +1,70 @@
...
+})
+export class RecordCardComponent implements OnInit {
+
+  @Input() record: Record;
+  @Input() attributesSortOrder: AttributeSort;
+  @Input() index: number;
+  @Input() extendedRecords: ExtendedRecord;
+  @Input() similarSearchHelper: any;
+  @Input() recordComments: RecordCommentsModal;
+  @Input() attributes: Map<string, any>;
```
JWittmeyer: if possible let's distance us from Maps a bit (I was the one introducing them however since they use functions (get) in the HTML this might not be optimal). Since we often use strings (uuids/attribute names) as keys they should be interchangeable. Objects keys etc. are more complex but often not necessary.

Also I'm not the biggest fan of the sortOrder/dict combination (Again something I introduced at some point I think) -> I think we should decide on a common approach for attributes throughout the UI.
A: double structure (dict+array+filter of array) -> benefit performance since only a small part is changed + dict access via key is fast
B: only array -> benefit redability -> will be replaced more often (maybe performance hit?) also maybe not since we need to loop anyway?

-> so maybe a global function to prepare an array? (again something we can talk about :) )

src/app/base/components/record-card/record-card.component.ts
```diff
@@ -0,0 +1,70 @@
...
+  storePreliminaryRecordIds(index: number) {
+    this.preliminaryRecordIds.emit(index);
+  }
+
+  requestSimilarSearch() {
+    const saveSimilaritySearch = this.dataBrowserModals.similaritySearch;
+    this.similarSearchHelper.requestSimilarSearch(saveSimilaritySearch.embeddingId, saveSimilaritySearch.recordId);
+  }
+
+  setEmbeddingIdSS(selectedValue: string) {
```
JWittmeyer: is the name of the function correct?

src/app/base/components/record-card/record-card.component.ts
lumburovskalina: That's true, no need for emitting, moved the logic from the parent component

src/app/base/components/record-card/components/record-display/record-display.component.ts
```diff
@@ -0,0 +1,38 @@
...
+        if (this.record.hasOwnProperty('recordData')) {
+          this.record.data = this.record.recordData;
+        }
+      }
+    }
+    if (changes.recordCardOptions) {
+      const firstEl = this.recordCardOptions.attributesSortOrder[0];
+      if (!firstEl.hasOwnProperty('key')) {
+        this.recordCardOptions.attributesSortOrder.forEach((attribute, index) => {
+          attribute.key = attribute.id;
```
JWittmeyer: if the provided structure doesnt have an id we should throw an error here imo (so check if exists if yes then use it otherwise error)

src/app/base/components/record-card/components/record-display/record-display.component.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  constructor() { }
+
+  ngOnChanges(changes: SimpleChanges): void {
+    if (changes.record) {
+      if (!this.record.hasOwnProperty('data')) {
+        if (this.record.hasOwnProperty('fullRecordData')) {
+          this.record.data = this.record.fullRecordData;
+        }
+        if (this.record.hasOwnProperty('recordData')) {
```
JWittmeyer: shouldn't this be else if? or is this intentionally overwritten when both exist?

JWittmeyer: also an else if neither exists with an error to ensure the developer uses it right

src/app/base/components/record-card/record-card-helper.ts
```diff
@@ -0,0 +1,29 @@
...
+import { SimilarSearch } from "src/app/data/components/data-browser/helper-classes/search-similar";
+import { ActiveSearchParams, AttributeSort, DataBrowserRecord } from "./record-card.types";
+
+export type RecordCardOptions = {
+    /**
+     * Options set for kern record card
+     * @attributesSortOrder {AttributeSort[] | any[], optional} - Array of sorted attributes to be displayed
+     * @dataBrowserModals {DataBrowserModals, optional} - Data browser modals with specific fields
+     * @indexExtendedRecord {number, optional} - Index of the extended record in the record list
+     * @recordList {DataBrowserRecord[], optional} - Array of records to be displayed
```
JWittmeyer: this doesn't make sense to me, why does the record card (singular instance) need a reference to the whole list?
As discussed before we should look for the stuff we explicitly need and only use that. For this instance I think an output event for clicked index should do it (the stage of the record ids would then be handled by the databrowser though)

src/app/base/components/record-card/components/record-display/record-display.component.ts
lumburovskalina: Yes it should be else if (I think either recordData or fullRecordData can be replacement for data but not both of them at the same time)

src/app/base/components/record-card/record-card-helper.ts
```diff
@@ -0,0 +1,29 @@
+import { DataBrowserModals, RecordCommentsModal } from "src/app/data/components/data-browser/helper-classes/modals-helper";
+import { SimilarSearch } from "src/app/data/components/data-browser/helper-classes/search-similar";
+import { ActiveSearchParams, AttributeSort, DataBrowserRecord } from "./record-card.types";
+
+export type RecordCardOptions = {
+    /**
+     * Options set for kern record card
+     * @attributesSortOrder {AttributeSort[] | any[], optional} - Array of sorted attributes to be displayed
+     * @dataBrowserModals {DataBrowserModals, optional} - Data browser modals with specific fields
+     * @indexExtendedRecord {number, optional} - Index of the extended record in the record list
```
JWittmeyer: i think just calling it recordIndex or currentRecordIndex would be more intuitive since the extendedRecord stuff is only loosely linked to the card component (since its located outside the folder scope)

JWittmeyer:
Lets create a record-list component inside the data browser module

- [x] resolved

src/app/data/components/record-list/record-list.component.ts
```diff
@@ -0,0 +1,65 @@
...
+  templateUrl: './record-list.component.html',
+  styleUrls: ['./record-list.component.scss']
+})
+export class RecordListComponent implements OnInit {
+
+  @Input() recordList: any[];
+  @Input() attributes: Attributes;
+  @Input() recordListOptions: RecordListOptions;
+
+  @Output() indexHuddleData = new EventEmitter<number>();
```
JWittmeyer: the name suggest to me that this is a number not an eventemmitter for a number. Maybe to generalize it a bit `recordClicked`?

src/app/data/components/record-list/record-list.component.ts
```diff
@@ -0,0 +1,65 @@
...
+  @Output() similarSearch = new EventEmitter();
```
JWittmeyer: maybe `similaritySearchRequested` ?

src/app/data/components/record-list/record-list.component.ts
```diff
@@ -0,0 +1,65 @@
...
+      displayName: 'Amount',
+      order: 4
+    });
+    columnsData.push({
+      field: 'confidenceAvg',
+      displayName: 'Avg.confidence',
+      order: 5
+    });
+    return columnsData;
+  }
```
JWittmeyer: if this is fixed anyway why dont be use a constant reference which is initiialized from the get go?
something like this outside the class: 
```
const DATA_BROWSER_TABLE_COLUMN_HEADERS:ColumnData[] = [
  { field: 'type', displayName: 'Type', order: 1 },
  { field: 'task', displayName: 'Task', order: 2 },
  { field: 'label', displayName: 'Label', order: 3 },
  { field: 'amount', displayName: 'Amount', order: 4 },
  { field: 'confidenceAvg', displayName: 'Avg.confidence', order: 5 }
]
```

and then just reference it in the onInit.

With this we wouldn't need to recrate it every time the browser is opened

JWittmeyer:
looks good. After the comments can be merged :)

JWittmeyer (approved):

PR #105, opened 2023-01-24
Data browser fixes

### LINKED ISSUE #198: [UX] - Incorrect wording in attribute calculation warning modal
**What would you improve?**
The modal currently says: "Be aware that calculated **records** are immutable.", which is incorrect.
It should instead say: "Be aware that calculated **attributes** are immutable."

image

### LINKED ISSUE #199: [BUG] - Outlier and Static slice view reset does not work in data browser
**Describe the bug**
When clearing the filter in the data browser on a static or outlier slice, the view does not reset properly and keeps showing the filtered data.

**To Reproduce**
1. Go to data browser
2. Click on outlier or static slice
3. Click "clear filter"
4. Observe that nothing happens

**Expected behavior**
When clearing the filter in the data browser on a static or outlier slice, the view should reset properly to show all records.

**Screenshots**
filter_reset_bug

 - OS: Win11 
 - Browser edge
 - Version 1.7.1

**Additional context**
-

### LINKED ISSUE #201: [UX] - Tooltip explanations for data browser filters disappear behind sidebar for some screens
**What would you improve?**
Currently for some screens, the tooltip explanations for attribute filters disappear when the operator is below the attribute selection. When they are side by side, it works fine. But from the screenshot you can see that I haven't even chosen a weird zoom setting. This was on my 1080p monitor with windows zoom setting set to 100%.

Solution could be to **a)** display it to the right, or **b)** display it over the sidebar.
I personally like **a)** more I think

image

### LINKED ISSUE #202: [UX] - Display of selected filters changes when changing the "connect by" option in the data browser
**What would you improve?**
This is a **really nitpicky issue**, but I've seen that the wording of the displayed filter option changes when switching the "connect by" option, which it should not need to do. It is a little irritating because your spots that something changed, but you cannot see what exactly because semantically nothing changed.

E.g. it changes from `running_id BEGINS WITH 1,22222122` to `running_id BEGINS_WITH 1,22222122` (see attached GIF for another example where there are also parentheses disappearing)

filter wording incorrect

### LINKED ISSUE #203: [UX] - Allow the usage of arrow keys when filtering for an integer attribute
**What would you improve?**
I cannot use arrow keys to reposition my cursor in the value field of the attribute filter when filtering for an integer attribute, which is annoying because I might want to add a wildcard before the current number and I have to delete everything, then insert the wildcard, and then re-enter everything.

image

### LINKED ISSUE #204: [BUG] - Filtering on data browser when an integer attribute is equal to 0 does not work
**Describe the bug**
Filtering on data browser with an integer attribute equal to 0 does not work. If the users choses an integer attribute from the dropdown menu and uses any operator except for IN and IN_WC and sets the search text to 0 , the filtering results are wrong because all records are returned. If the user choses any other search text or non integer attribute, this wouldn't be the case.

**To Reproduce**
1. Go to data-browser.
2. Select integer attribute(ex. running_id), operator EQUAL and type 0 into the search text.
3. The filtering gives wrong results.

**Expected behavior**
The expected behaviour is that the user gets the right data (ex. if he searches for attribute EQUAL to zero, he should get only 0 result and not all records). This should work for all operators, different data types and different search text.

**Screenshots**
Screenshot 2023-01-11 at 16 22 38

 - OS: [macOS]
 - Browser [chrome]

### LINKED ISSUE #210: [BUG] - Multiple attribute filters to not combine properly
**Describe the bug**
When filtering for more than one attribute filter, it should display the intersection of it. Currently this is not the case (see screenshot). My suspicion is that this has something to do with the "any attribute" filter setting. If condition it on actual attributes I cannot create such an inconsistent state, but am not 100% sure.

Also, if I search for a text in the "any attribute" then this bug does not happen, only if I search for a number (see GIF). There might be other combinations with different operators where this fails, though. But that might be valuable information for bug identification.

**To Reproduce**
1. Go to data browser in AG News project
2. Create attribute filter with "running_id" EQUAL 501
3. Create another attribute filter with "any attribute" CONTAINS 7

**Expected behavior**
There should be no record displayed as both conditions at the same time are not true on any record

**Screenshots**
image
data_browser_filter_multiple_bug

 - OS: Win11
 - Browser edge
 - Version 1.7.1
 
**Additional context**
-

### LINKED ISSUE #211: [UX] - Filter wording of "ML-label" as "manual label" could be confusing
**What would you improve?**
Currently, when you set a manual label filter in the data browser, you get the abbreviation "ML-label" displayed. As "ML" is a common abbreviation for Machine Learning, this might be rather confusing. I would just lose the "L" and call it "M-label".

image

### LINKED ISSUE #212: [UX] - The open-source version displays options to view data slice as expert, which has no functionality
**What would you improve?**
When you click on the information icon on the data slices, you can click "view as expert" in the open-source version, which brings you to an expert view, where you cannot do anything. As the open-source version does not have user management, I suggest disabling this button with the tooltip information that this is only relevant for the managed version with user management capabilities.

image

### LINKED ISSUE #213: [BUG] - Missing UI reset after clearing filters in data browser
**Describe the bug**
After deselecting a data slice that has ordering in the filter settings, the ordering setting in the UI is not reset even though it is not applied.

**To Reproduce**
1. Go to data browser
2. Make data slice with ordering
3. select that slice
4. press "clear filter"
5. observe that ordering is still in the UI in filter settings but not applied anymore

**Expected behavior**
The ordering setting should reset.

**Screenshots**
filter_reset

 - OS: Win10
 - Browser Chrome
 - Version 1.7.1

**Additional context**
-

lumburovskalina:
Implements/fixes:
- closes https://github.com/code-kern-ai/refinery/issues/211
- closes https://github.com/code-kern-ai/refinery/issues/212
- closes https://github.com/code-kern-ai/refinery/issues/199
- closes https://github.com/code-kern-ai/refinery/issues/203
- closes https://github.com/code-kern-ai/refinery/issues/202
- closes https://github.com/code-kern-ai/refinery/issues/204
- closes https://github.com/code-kern-ai/refinery/issues/213
- closes https://github.com/code-kern-ai/refinery/issues/201
- closes https://github.com/code-kern-ai/refinery/issues/210
- closes https://github.com/code-kern-ai/refinery/issues/198

PRs:
- https://github.com/code-kern-ai/refinery-ui/pull/105
- https://github.com/code-kern-ai/refinery-gateway/pull/100


JWittmeyer:
not sure why but now when i enter the databrowser page it loads twice (two requests in network tab and the record display flickers once)

- [ ] resolved

JWittmeyer:
not sure why but now when i enter the databrowser page it loads twice (two requests in network tab and the record display flickers once)

- [x] resolved

src/app/data/components/data-browser/data-browser.component.html
```diff
@@ -969,8 +974,10 @@ <h1 class="text-lg text-gray-900 mb-2 text-center">Info</h1>
                     {{item.value.substring(item.value.startsWith("https")?8:7)}}</span>
 
             </div>
-            <div>
-                <button (click)="testLink(item.value + '?pos=1&type=DATA_SLICE')"
+            <div [attr.data-tip]="!isManaged ? 'Only usable for the managed version' : ''"
```
JWittmeyer: doesn't need to be bound since having it without the tooltip class doesn't do anything :)

JWittmeyer (approved):

PR #109, opened 2023-01-30
Scrolling before animation finishes fix on attributes and heuristics pages

### LINKED ISSUE #209: [UX] - Scrolling before animation finished causes inconsistent header bar in attribute calculation
**What would you improve?**
When you are in the attribute calculation screen and scroll down there is a nice animation on the top of the screen for the header bar that lets you go back and displays the attribute name and state. If you scroll back up there is also an animation that puts the info back in place. But if you scroll up and then down again before that animation has finished, you get different - slightly uglier - header bar. I don't know how all of this works but it would be great if the header bar was consistent no matter how weird the user's scrolling behavior might be.

Tested on Win10 and Edge browser.

scrolling_through_animation

lumburovskalina:
Implements:
- closes https://github.com/code-kern-ai/refinery/issues/209

PRs:


JWittmeyer:
issue still occurs on my side 

Idea proposal: On scroll event start timer that resets itself on scroll event (maybe ~500ms) after that timer we check the page div scroll position if at pos > X then version 1 of the header otherwise version 2.

JWittmeyer (approved):

PR #110, opened 2023-01-30
Attribute name in the monaco editor fix when using brics

### LINKED ISSUE #218: [BUG] - Attribute calculation function name in monaco editor
**Describe the bug**
When creating a new attribute the function code in the monaco editor holds ac instead of the function name. A short switch can be noticed

**To Reproduce**
1. Go to settings
2. Click on new attribute (e.g. text)
3. Use a brick (e.g. language detection)
4. See the function name 

Also noticeable it seems to run without any issues however it looks a bit odd having the name of the attribute diverging form the one in the code window.

**Expected behavior**
Same name with the attribute

**Screenshots**
image

 - OS: Windows
 - Browser chrome
 - Version 1.7.1 (dev so maybe 1.8.0)

**Additional context**
Happened during streaming of new platform on Jens pc -- so maybe also performance/platform related (though very unlikely)

lumburovskalina:
Implements:
- Closes https://github.com/code-kern-ai/refinery/issues/218

PRs:

### DIFF
--- a/src/app/base/components/bricks-integrator/bricks-integrator.component.ts
+++ b/src/app/base/components/bricks-integrator/bricks-integrator.component.ts
@@ -39,7 +39,6 @@ export class BricksIntegratorComponent implements OnInit, OnDestroy {
   @Output() newTaskId = new EventEmitter<string>();
 
   @ViewChild("searchInput") searchInput: ElementRef;
-  @ViewChild("functionNameInput") functionNameInput: ElementRef;
 
   config: BricksIntegratorConfig;
   codeParser: BricksCodeParser;
@@ -180,7 +179,6 @@ export class BricksIntegratorComponent implements OnInit, OnDestroy {
         case IntegratorPage.INPUT_EXAMPLE:
           // jump to integration
           this.config.page = IntegratorPage.INTEGRATION;
-          timer(100).subscribe(() => this.functionNameInput.nativeElement.focus());
           break;
         case IntegratorPage.INTEGRATION:
           //transfer code to editor
@@ -196,9 +194,6 @@ export class BricksIntegratorComponent implements OnInit, OnDestroy {
     if (page == IntegratorPage.SEARCH || (this.config.api.requesting || this.config.api.data)) {
       this.config.page = page;
       this.checkCanAccept();
-    }
-    if (page == IntegratorPage.INTEGRATION) {
-      timer(100).subscribe(() => this.functionNameInput.nativeElement.focus());
     }
   }
 
--- a/src/app/base/components/bricks-integrator/bricks-integrator.component.ts
+++ b/src/app/base/components/bricks-integrator/bricks-integrator.component.ts
@@ -349,7 +349,7 @@ export class BricksIntegratorComponent implements OnInit, OnDestroy {
     this.checkCanAccept();
     if (this.config.preparedCode) {
       this.codeParser.functionName = name;
-      this.codeParser.baseCode = this.codeParser.baseCode.replace(getPythonFunctionName(this.config.preparedCode), name);
+      this.config.preparedCode = this.config.preparedCode.replace(getPythonFunctionName(this.config.preparedCode), name);
     }
   }
 }
--- a/src/app/base/components/bricks-integrator/bricks-integrator.component.ts
+++ b/src/app/base/components/bricks-integrator/bricks-integrator.component.ts
@@ -1,7 +1,7 @@
 
 import { HttpClient } from '@angular/common/http';
 import { Component, ElementRef, EventEmitter, Input, OnChanges, OnDestroy, OnInit, Output, SimpleChanges, ViewChild } from '@angular/core';
-import { ActivatedRoute } from '@angular/router';
+import { ActivatedRoute, Router } from '@angular/router';
 import { timer } from 'rxjs';
 import { first } from 'rxjs/operators';
 import { findProjectIdFromRoute, getPythonFunctionName, isStringTrue, toPythonFunctionName } from 'src/app/util/helper-functions';
@@ -44,9 +44,10 @@ export class BricksIntegratorComponent implements OnInit, OnDestroy {
   config: BricksIntegratorConfig;
   codeParser: BricksCodeParser;
   dataRequestor: BricksDataRequestor;
+  functionType: string;
 
   constructor(private http: HttpClient, private projectApolloService: ProjectApolloService, private knowledgeBaseApollo: KnowledgeBasesApolloService,
-    private activatedRoute: ActivatedRoute,) {
+    private activatedRoute: ActivatedRoute, private router: Router) {
   }
 
   ngOnInit(): void {
@@ -55,6 +56,7 @@ export class BricksIntegratorComponent implements OnInit, OnDestroy {
     this.codeParser = new BricksCodeParser(this);
     const projectId = findProjectIdFromRoute(this.activatedRoute)
     this.dataRequestor = new BricksDataRequestor(this.projectApolloService, this.knowledgeBaseApollo, projectId, this);
+    this.functionType = this.router.url.indexOf("attributes") > -1 ? "Attribute" : this.router.url.indexOf("heuristics") > -1 ? "Heuristic" : "Function";
   }
   ngOnDestroy(): void {
     this.dataRequestor.unsubscribeFromWebsocket();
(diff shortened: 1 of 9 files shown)


src/app/base/components/bricks-integrator/helper/code-parser.ts
```diff
@@ -47,7 +49,7 @@ export class BricksCodeParser {
     }
 
     public replaceVariables() {
-        let replacedCode = this.baseCode;
+        let replacedCode = this.baseCode.replace(getPythonFunctionName(this.baseCode), this.functionName);
```
JWittmeyer: not sure if this approach work reliably since only the functionname is used to replace. This means if the function name is e.g. provided in a previous comment this will run into issues. I think we should try to find the line index of the function and change the value at that index to be sure.

src/app/base/components/bricks-integrator/bricks-integrator.component.ts
```diff
@@ -340,5 +344,13 @@ export class BricksIntegratorComponent implements OnInit, OnDestroy {
     if (this.newTaskId.observers.length > 0) this.newTaskId.emit(taskId);
   }
 
-
+  checkIfFunctionNameIsUnique(name: string) {
+    this.codeParser.nameTaken = this.nameLookups?.find(x => x == name) != undefined;
```
JWittmeyer: !!(this.nameLookups?.find(x => x == name))

JWittmeyer:
branch name with _ instead of -

src/app/base/components/bricks-integrator/bricks-integrator.component.ts
```diff
@@ -340,5 +344,13 @@ export class BricksIntegratorComponent implements OnInit, OnDestroy {
...
+    name = toPythonFunctionName(name);
+    this.checkCanAccept();
+    if (this.config.preparedCode) {
+      this.codeParser.functionName = name;
+      this.config.preparedCode = this.config.preparedCode.replace(getPythonFunctionName(this.config.preparedCode), name);
```
JWittmeyer: shouldn't work as expected because of replace issues. Also, the canAccept variable only seems to be set in replaceVariables function which isn't called here (which is correct). But the accept state isn't checked so we should create a function to check the state and call it

JWittmeyer:
Funciton name input should be part of the same grid as the other parts
image

- [x] resolved

JWittmeyer:
Function name input looks different to other inputs

- [x] resolved

JWittmeyer:
I think we should change the naming to something more clear for the user. its a function name but it will be the xxx name (attribute or heuristic).
image

- [x] resolved

src/app/base/components/bricks-integrator/bricks-integrator.component.ts
JWittmeyer: ok didn't read fully. The checkCanAccept function is the correct one. Also it should be run at the end imo.

JWittmeyer:
Function name input still different:
image
- focus effect should be removed
- col span only 1 
- to put it a bit more in line with the other gird entries I would put the function name thing as alert triangle from table icons.
-> name is free => nothing is shown in 3rd grid column
-> name isn't free => red alert triangle with data tip "XXXX name exists". since we don't change the type after creation but on initialization i'd go with string interpolation instead of binding the attribute so: `attr.data-tip="{{functionType + ' name exists'}}"`

- [x] resolved

src/app/base/components/bricks-integrator/bricks-integrator.component.ts
```diff
@@ -340,5 +346,13 @@ export class BricksIntegratorComponent implements OnInit, OnDestroy {
     if (this.newTaskId.observers.length > 0) this.newTaskId.emit(taskId);
   }
 
-
+  checkIfFunctionNameIsUnique(name: string) {
+    this.codeParser.nameTaken = !!(this.nameLookups?.find(x => x == name));
+    name = toPythonFunctionName(name);
+    if (this.config.preparedCode) {
+      this.codeParser.functionName = name;
+      this.config.preparedCode = this.config.preparedCode.replace(new RegExp(getPythonFunctionName(this.config.preparedCode), 'g'), name);
```
JWittmeyer: this wasn't meant with the previous comment. I mean that a simple replace of the name might run into issues independent of global or not (even more so with global but in general not viable).  
image


Also this doesn't really belong in the component ts file. 
I'd go with a function in the codeparser that is used to set the function name that in turn handles the preparedCode. 

This should also ensure that we always start from scratch in the preparation. I want to remove the possibility that someone changes a function name and changes it again and arrives at a different result than selecting the final name from the beginning.
button



Also the function name might be misleading since it's more of a checkFunctionNameAndSet or something similar since it not only to check

src/app/base/components/bricks-integrator/helper/code-parser.ts
```diff
@@ -48,6 +50,11 @@ export class BricksCodeParser {
 
     public replaceVariables() {
         let replacedCode = this.baseCode;
+        const idxReplace = this.getIndexFunctionLine();
```
JWittmeyer: should be checked for -1 to stop if none was found.

src/app/base/components/bricks-integrator/helper/code-parser.ts
```diff
@@ -327,4 +356,25 @@ export class BricksCodeParser {
...
         }
     }
+
+    checkFunctionNameAndSet(name: string) {
+        if (!name) return;
+        this.nameTaken = !!(this.base.nameLookups?.find(x => x == name));
+        name = toPythonFunctionName(name);
+        if (this.base.config.preparedCode) {
+            this.functionName = name;
+            this.base.config.preparedCode = this.replaceFunctionLine(this.base.config.preparedCode);
```
JWittmeyer: this doesn't start from scratch as discussed. For questions reach out

src/app/base/components/bricks-integrator/bricks-integrator.component.ts
```diff
@@ -55,6 +55,7 @@ export class BricksIntegratorComponent implements OnInit, OnDestroy {
     this.codeParser = new BricksCodeParser(this);
     const projectId = findProjectIdFromRoute(this.activatedRoute)
     this.dataRequestor = new BricksDataRequestor(this.projectApolloService, this.knowledgeBaseApollo, projectId, this);
+    this.functionType = this.router.url.indexOf("attributes") > -1 ? "Attribute" : this.router.url.indexOf("heuristics") > -1 ? "Heuristic" : "Function";
```
JWittmeyer: i think the url is a bit unreliable here. especially since active learner (classes) and python functions have the same URL (besides a parameter). I'd go with an input parameter to be set for the component.

src/app/base/components/bricks-integrator/helper/code-parser.ts
```diff
@@ -147,6 +153,37 @@ export class BricksCodeParser {
     }
 
 
+    getCurrentFunctionLine(): string {
+        const lines = this.base.config.preparedCode.split("\n");
+        for (let i = 0; i < lines.length; i++) {
+            const line = lines[i].trim();
+            if (line.startsWith('def ' + this.functionName + '(record')) {
+                return line;
+            }
```
JWittmeyer: this one doesn't seem to differentiate between active lerner and function?

JWittmeyer:
Empty Function names seem to work
image

- [x] resolved

JWittmeyer:
- focus effect should be removed
- changes take a bit to take effect. I assume it's because of (keyup). E.g. this results in jumping to the end for a space in the middle and holding down e.g. space results in only one _ 

Tested a bit around and found this to be working good enough:
Function:

  ``` Typescript
onInputFunctionName(event: Event) {
    if (!(event.target instanceof HTMLInputElement)) return;
    const start = event.target.selectionStart;
    let value = event.target.value;
    this.codeParser.checkFunctionNameAndSet(value)
    event.target.value = this.codeParser.functionName;
    event.target.selectionStart = start;
    event.target.selectionEnd = start;
  }
```
HTML:
``` HTML
<input class="input input-bordered col-span-1 input-sm" type="text"
    style="outline:none; box-shadow:none;" (input)="onInputFunctionName($event)"
    [value]="codeParser.functionName" />
```


lock

- [x] resolved

JWittmeyer (approved):

PR #111, opened 2023-02-01
removes imports but not package

JWittmeyer:


src/app/app.component.ts
```diff
@@ -81,34 +79,10 @@ export class AppComponent implements OnDestroy, OnInit, AfterViewInit {
...
-    this.initializeIntercom();
     //caution! the first user request needs to run after the db creation since otherwise the backend will try to create an unasigned user
     this.organizationService.getUserInfo().pipe(first()).subscribe((v) => UserManager.initUserManager(this.router, this.organizationService, v));
   }
 
-  initializeIntercom() {
-
-    const token = ConfigManager.getConfigValue("tokens", "INTERCOM")
-
-    this.userSubscription$ = this.organizationService
```
lumburovskalina, line 93: Variable `userSubscription$` is not used anymore and can be removed. The same goes for `loggedInUser` since its value is never set

lumburovskalina (approved):

PR #114, opened 2023-02-06
Pre-filling of active learner embedding selection fix

### LINKED ISSUE #200: [UX] - Data slices have different names throughout the application
**What would you improve?**
We have: "filter", "slice", and "data slice" in our application currently. Maybe some more variations that I missed. I personally like "data slice" or "slice" the most, even more than "filter".

I think we should unify this naming, also for consistency in the docs.

image
image
image

### LINKED ISSUE #219: [BUG] - data browser filter for empty doesn't work
**Describe the bug**
filtering for equal empty return all records an not just the ones with no entry

**To Reproduce**
1. Go to data browser
2. Expand attributes
3. Change filter to equal
4. (Any attribute equal '')
5. all records are shown

**Expected behavior**
only show the ones with at least one attribute none existent

**Screenshots**
image

 - OS: Windows
 - Browser chrome
 - Version 1.7.1 dev (so more like 1.8.0)

**Additional context**
happens for both any attribute and specific attribute selected

### LINKED ISSUE #221: [BUG] - Lookup lists that are not valid Python variable names break the labeling function calculation
**Describe the bug**
Lookup lists that are not valid variable names break the labeling function calculation even when they are not used. When I removed the list called "123" (see screenshot), the calculation worked again. 

**To Reproduce**
1. Create a lookup list called "123"
2. Import some list (dsnt have to be the 123 one) from knowledge in a LF
3. Run it

**Expected behavior**
Either I shouldnt be able to create such list names or the execution environment should preface the list name with something viable like "list_"

**Screenshots**
image

 - OS: win10
 - Browser chrome
 - Version 1.7.1

**Additional context**
-

### LINKED ISSUE #222: [BUG] - Pre-filling of Active learner embedding selection selects the wrong embedding
**Describe the bug**
When selecting embedding B in the active learner creation modal, it will not be prefilled if there is another embedding A that comes before it. Dangerous because this is rather silent and needs user attention to verify.

**To Reproduce**
1. have two embeddings for an attribute
2. create new AL
3. select second embedding for AL creation
4. See error in code

**Expected behavior**
correct embedding should be selected

**Screenshots**
albug

remote dev environment, 1.7.1

**Additional context**
-

lumburovskalina:
Implements:
- Closes https://github.com/code-kern-ai/refinery/issues/222
- Closes https://github.com/code-kern-ai/refinery/issues/219
- Closes https://github.com/code-kern-ai/refinery/issues/221
- Closes https://github.com/code-kern-ai/refinery/issues/200

PRs:
- https://github.com/code-kern-ai/refinery-ui/pull/114
- https://github.com/code-kern-ai/refinery-gateway/pull/105

### DIFF
--- a/src/app/base/components/export/export-helper.ts
+++ b/src/app/base/components/export/export-helper.ts
@@ -24,7 +24,7 @@ export enum ExportPreset {
 
 export enum ExportRowType {
     ALL = "ALL",
-    SLICE = "SLICE",
+    SLICE = "DATA_SLICE",
     SESSION = "SESSION"
 }
 
--- a/src/app/base/components/export/export-helper.ts
+++ b/src/app/base/components/export/export-helper.ts
@@ -24,7 +24,7 @@ export enum ExportPreset {
 
 export enum ExportRowType {
     ALL = "ALL",
-    SLICE = "DATA_SLICE",
+    DATA_SLICE = "DATA_SLICE",
     SESSION = "SESSION"
 }
 
@@ -110,7 +110,7 @@ export class ExportHelper {
         let type = this.firstActiveInGroup(ExportEnums.ExportRowType, 'value');
         let id;
         if (type == ExportRowType.ALL) id = null;
-        else if (type == ExportRowType.SLICE) {
+        else if (type == ExportRowType.DATA_SLICE) {
             id = this.firstActiveInGroup(ExportEnums.DataSlices, "id");
         } else if (type == ExportRowType.SESSION) id = this.baseComponent.sessionId;
         return { type: type, id: id };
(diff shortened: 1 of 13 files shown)


src/app/base/components/export/export-helper.ts
```diff
@@ -24,7 +24,7 @@ export enum ExportPreset {
 
 export enum ExportRowType {
     ALL = "ALL",
-    SLICE = "SLICE",
+    SLICE = "DATA_SLICE",
```
JWittmeyer: somewhat dangerous, we should change the enum entry as well (SLICE = "DATA_SLICE" to DATA_SLICE = "DATA_SLICE")

JWittmeyer:
extraction label issue - lookuplists can have any name, the python variable now (needs backendchange in variable name resolveing)
image


- [x] resolved

JWittmeyer (approved):

PR #117, opened 2023-02-08
Close modal without saving the data slice and visual improvements on data-browser

lumburovskalina:
- Close modal on data browser does not save the data slice
- Visual improvements for icon for outdated slice
- Tooltip class removed from all files and added into the global styles
- Fix for loading icon on embeddings subsection when there are no text attributes


JWittmeyer:
Outdated jumps a bit to much for me:
lock

I'd put it  in the line next to filter if that's easily doable (don't know because of the -----)
image

- [x] resolved

JWittmeyer (approved):

PR #137, opened 2023-05-11
Progress bar on LF and AC

lumburovskalina:
PRs:
- https://github.com/code-kern-ai/refinery-ui/pull/137
- https://github.com/code-kern-ai/refinery-gateway/pull/129
- https://github.com/code-kern-ai/refinery-submodule-model/pull/42
- https://github.com/code-kern-ai/refinery-ac-exec-env/pull/19

### DIFF
--- a/src/app/projects/components/create-new-attribute/create-new-attribute.component.html
+++ b/src/app/projects/components/create-new-attribute/create-new-attribute.component.html
@@ -278,28 +278,6 @@
                                     class="tooltip tooltip-right relative z-10">
                                     <kern-loading></kern-loading>
                                 </div>
-                                <div *ngIf="currentAttribute.state === 'USABLE'" data-tip="Successfully executed."
-                                    class="tooltip tooltip-right">
-                                    <div class="btn cursor-default btn-circle btn-xs no-animation btn-primary mx-2">
-                                        <svg xmlns="http://www.w3.org/2000/svg"
-                                            class="inline-block w-4 h-4 stroke-current" fill="none" viewBox="0 0 24 24"
-                                            stroke="currentColor">
-                                            <path stroke-linecap="round" stroke-linejoin="round" stroke-width="2"
-                                                d="M5 13l4 4L19 7" />
-                                        </svg>
-                                    </div>
-                                </div>
-                                <div *ngIf="currentAttribute.state === 'ERROR'" data-tip="Execution ran into errors."
-                                    class="tooltip tooltip-right">
-                                    <div class="btn cursor-default btn-circle btn-xs no-animation btn-error mx-2">
-                                        <svg xmlns="http://www.w3.org/2000/svg"
-                                            class="inline-block w-4 h-4 stroke-current" fill="none" viewBox="0 0 24 24"
-                                            stroke="currentColor">
-                                            <path stroke-linecap="round" stroke-linejoin="round" stroke-width="2"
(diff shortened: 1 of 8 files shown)


JWittmeyer:
why is this branch name `progress-bar-lf-ac` and not e.g. `ac-progress-bar` (or vice versa)?

- [x] no need to change but I acknowledge that linked repository PRs should use the same branch name

src/app/projects/components/create-new-attribute/create-new-attribute.component.html
```diff
@@ -263,9 +263,73 @@
 
 
             <kern-logs [logs]="attributeLogs"></kern-logs>
+
+            <div class="mt-8">
+                <div class="text-sm leading-5 font-medium text-gray-700 inline-block">Calculation progress</div>
+                <div *ngIf="currentAttribute.progress == 0 || currentAttribute.state == 'INITIAL'" class="bg-white">
+                    <div class="py-6 text-sm leading-5 font-normal text-gray-500">This attribute was not yet run.</div>
+                </div>
+                <div *ngIf="currentAttribute.progress > 0 && currentAttribute.progress < 1"
```
JWittmeyer: i think we should also display on progress  = 0 (but state running or something like that). Since the progress bar now "blinks in" after some calculation is already run

src/app/projects/components/create-new-attribute/create-new-attribute.component.html
```diff
@@ -263,9 +263,73 @@
...
+                        </div>
+                        <ng-template [ngIf]="currentAttribute.state === 'RUNNING'">
+                            <div class="text-sm leading-5 font-normal text-gray-500">
+                                <div class="flex flex-row justify-between items-center">
+                                    <div class="flex flex-grow items-center">
+                                        <div class="w-full bg-gray-200 rounded-full dark:bg-gray-700 h-3">
+                                            <div class="bg-blue-600 font-extra-small font-medium text-blue-100 h-3 text-center leading-none rounded-full"
+                                                [ngStyle]="{'width': (currentAttribute.progress * 100) + '%'}">
+                                                {{currentAttribute.progress * 100}}%</div>
+                                        </div>
```
JWittmeyer: why is this used and not a common `<progress>` element?

lumburovskalina: For the labeling functions we use the number of iterations besides the progress bar itself but for the ac we do not have that (if it is successful it is run only once), so I thought adding the number of percentage as a label would be a good idea. Otherwise, imo the bar looked 'empty'. Adding the label on the <progress> element had some issues and I saw that on the upload component we use this approach (one bar is with the progress element and one is with this approach). I think changing the <progress> element with Tailwind classes is a good idea to get rid of daisyUI. 
Not sure if we want to replace it everywhere now?

src/app/projects/components/create-new-attribute/create-new-attribute.component.ts
```diff
@@ -324,6 +325,9 @@ export class CreateNewAttributeComponent implements OnInit, OnDestroy {
       this.updatedThroughWebsocket = true;
       this.currentAttributeQuery$.refetch();
     } else if (msgParts[1] == 'calculate_attribute') {
+      if (msgParts[2] == 'progress') {
+        this.currentAttribute.progress = Number(msgParts[4]);
+      }
       this.attributesQuery$.refetch();
       this.currentAttributeQuery$.refetch();
```
JWittmeyer: i think we dont need both. If the progress is updated directly we don't need to refetch the whole thing. (at least I think so). If we do need to refetch then the progress update itself is unnecessary.

JWittmeyer (approved):

PR #143, opened 2023-06-02
Embedding providers

lumburovskalina:
gateway: https://github.com/code-kern-ai/refinery-gateway/pull/136
embedder: https://github.com/code-kern-ai/refinery-embedder/pull/42
gates: https://github.com/code-kern-ai/gates-runtime/pull/24
submodule model: https://github.com/code-kern-ai/refinery-submodule-model/pull/44
ui: https://github.com/code-kern-ai/refinery-ui/pull/143
admin-dashboard: https://github.com/code-kern-ai/admin-dashboard/pull/8
updater: https://github.com/code-kern-ai/refinery-updater/pull/43

### DIFF
--- a/src/app/base/services/project/project-apollo.service.ts
+++ b/src/app/base/services/project/project-apollo.service.ts
@@ -1489,4 +1489,17 @@ export class ProjectApolloService {
         })
       );
   }
+
+  getEmbeddingPlatforms() {
+    return this.apollo
+      .query({
+        query: queries.GET_EMBEDDING_PLATFORMS,
+        fetchPolicy: 'no-cache',
+      })
+      .pipe(
+        map((result) => {
+          return result['data']['embeddingPlatforms'];
+        })
+      );
+  }
 }
--- a/src/app/base/services/project/project-apollo.service.ts
+++ b/src/app/base/services/project/project-apollo.service.ts
@@ -737,13 +737,14 @@ export class ProjectApolloService {
       );
   }
 
-  createEmbedding(projectId: string, attributeId: string, embeddingHandle: string, granularity: string) {
+  createEmbedding(projectId: string, attributeId: string, embeddingHandle: string, granularity: string, platform: string) {
     return this.apollo.mutate({
       mutation: granularity == "TOKEN" ? mutations.CREATE_TOKEN_LEVEL_EMBEDDING : mutations.CREATE_ATTRIBUTE_LEVEL_EMBEDDING,
       variables: {
         projectId: projectId,
         attributeId: attributeId,
         embeddingHandle: embeddingHandle,
+        platform: platform,
       },
       refetchQueries: [
         {
--- a/src/app/base/services/project/project-apollo.service.ts
+++ b/src/app/base/services/project/project-apollo.service.ts
@@ -10,6 +10,7 @@ import { ApolloChecker } from '../base/apollo-checker';
 import { mutations } from './project-mutations';
 import { queries } from './project-queries';
 import { dateAsUTCDate } from 'submodules/javascript-functions/date-parser';
+import { EmbeddingPlatform } from 'src/app/projects/components/project-settings/entities/embedding.type';
 
 @Injectable({
   providedIn: 'root',
@@ -738,7 +739,7 @@ export class ProjectApolloService {
       );
   }
 
-  createEmbedding(projectId: string, attributeId: string, config: any) {
+  createEmbedding(projectId: string, attributeId: string, config: string) {
     return this.apollo.mutate({
       mutation: mutations.CREATE_EMBEDDING,
       variables: {
@@ -1490,7 +1491,7 @@ export class ProjectApolloService {
       );
   }
 
-  getEmbeddingPlatforms() {
+  getEmbeddingPlatforms(): Observable<EmbeddingPlatform> {
     return this.apollo
       .query({
         query: queries.GET_EMBEDDING_PLATFORMS,
--- a/src/app/base/services/project/project-apollo.service.ts
+++ b/src/app/base/services/project/project-apollo.service.ts
@@ -737,17 +737,13 @@ export class ProjectApolloService {
       );
   }
 
-  createEmbedding(projectId: string, attributeId: string, embeddingHandle: string, granularity: string, platform: string, model: string, apiToken: string, acceptTerms: boolean) {
+  createEmbedding(projectId: string, attributeId: string, granularity: string, newEmbedding: any) {
     return this.apollo.mutate({
       mutation: granularity == "TOKEN" ? mutations.CREATE_TOKEN_LEVEL_EMBEDDING : mutations.CREATE_ATTRIBUTE_LEVEL_EMBEDDING,
       variables: {
         projectId: projectId,
         attributeId: attributeId,
-        embeddingHandle: embeddingHandle,
-        platform: platform,
-        model: model,
-        apiToken: apiToken,
-        acceptTerms: acceptTerms,
+        newEmbedding: newEmbedding,
       },
       refetchQueries: [
         {
--- a/src/app/base/services/project/project-apollo.service.ts
+++ b/src/app/base/services/project/project-apollo.service.ts
@@ -737,9 +737,9 @@ export class ProjectApolloService {
       );
   }
 
-  createEmbedding(projectId: string, attributeId: string, granularity: string, newEmbedding: any) {
+  createEmbedding(projectId: string, attributeId: string, newEmbedding: any) {
     return this.apollo.mutate({
-      mutation: granularity == "TOKEN" ? mutations.CREATE_TOKEN_LEVEL_EMBEDDING : mutations.CREATE_ATTRIBUTE_LEVEL_EMBEDDING,
+      mutation: mutations.CREATE_EMBEDDING,
       variables: {
         projectId: projectId,
         attributeId: attributeId,
--- a/src/app/base/services/project/project-apollo.service.ts
+++ b/src/app/base/services/project/project-apollo.service.ts
@@ -1495,7 +1495,7 @@ export class ProjectApolloService {
     return this.apollo
       .query({
         query: queries.GET_EMBEDDING_PLATFORMS,
-        fetchPolicy: 'no-cache',
+        fetchPolicy: 'cache-first',
       })
       .pipe(
         map((result) => {
--- a/src/app/base/services/project/project-apollo.service.ts
+++ b/src/app/base/services/project/project-apollo.service.ts
@@ -737,7 +737,7 @@ export class ProjectApolloService {
       );
   }
 
-  createEmbedding(projectId: string, attributeId: string, embeddingHandle: string, granularity: string, platform: string) {
+  createEmbedding(projectId: string, attributeId: string, embeddingHandle: string, granularity: string, platform: string, model: string, apiToken: string, acceptTerms: boolean) {
     return this.apollo.mutate({
       mutation: granularity == "TOKEN" ? mutations.CREATE_TOKEN_LEVEL_EMBEDDING : mutations.CREATE_ATTRIBUTE_LEVEL_EMBEDDING,
       variables: {
@@ -745,6 +745,9 @@ export class ProjectApolloService {
         attributeId: attributeId,
         embeddingHandle: embeddingHandle,
         platform: platform,
+        model: model,
+        apiToken: apiToken,
+        acceptTerms: acceptTerms,
       },
       refetchQueries: [
         {
--- a/src/app/base/services/project/project-apollo.service.ts
+++ b/src/app/base/services/project/project-apollo.service.ts
@@ -737,13 +737,13 @@ export class ProjectApolloService {
       );
   }
 
-  createEmbedding(projectId: string, attributeId: string, newEmbedding: any) {
+  createEmbedding(projectId: string, attributeId: string, config: any) {
     return this.apollo.mutate({
       mutation: mutations.CREATE_EMBEDDING,
       variables: {
         projectId: projectId,
         attributeId: attributeId,
-        newEmbedding: newEmbedding,
+        config: config,
       },
       refetchQueries: [
         {
--- a/src/app/import/components/helpers/upload-specific.ts
+++ b/src/app/import/components/helpers/upload-specific.ts
@@ -26,7 +26,7 @@ export class RecordNewUploadHelper {
 
 export class RecordAddUploadHelper {
     projectName: string = '';
-    hasNotGdprEmbeddings: boolean = false;
+    recalculationCosts: boolean = false;
 
     constructor(private baseComponent: UploadComponent) {
         this.baseComponent = baseComponent;
--- a/src/app/import/components/helpers/upload-specific.ts
+++ b/src/app/import/components/helpers/upload-specific.ts
@@ -26,6 +26,7 @@ export class RecordNewUploadHelper {
 
 export class RecordAddUploadHelper {
     projectName: string = '';
+    hasGdprEmbeddings: boolean = false;
 
     constructor(private baseComponent: UploadComponent) {
         this.baseComponent = baseComponent;
--- a/src/app/import/components/helpers/upload-specific.ts
+++ b/src/app/import/components/helpers/upload-specific.ts
@@ -26,7 +26,7 @@ export class RecordNewUploadHelper {
 
 export class RecordAddUploadHelper {
     projectName: string = '';
-    hasGdprEmbeddings: boolean = false;
+    hasNotGdprEmbeddings: boolean = false;
 
     constructor(private baseComponent: UploadComponent) {
         this.baseComponent = baseComponent;
--- a/src/app/import/components/upload/upload.component.html
+++ b/src/app/import/components/upload/upload.component.html
@@ -262,7 +262,8 @@
 <ng-template #cautionMessage>
     <div class="text-sm text-gray-500 font-normal">
         <strong class="text-sm text-gray-700 font-medium">CAUTION:</strong> Submitting new records will automatically
-        run attribute calculation and embeddings for the newly
-        uploaded records
+        run attribute calculation and embeddings <span *ngIf="uploadSpecificHelper.hasGdpr">(recreation of the
+            embeddings will calculate them one more time)</span>
+        for the newly uploaded records
     </div>
 </ng-template>
--- a/src/app/import/components/upload/upload.component.html
+++ b/src/app/import/components/upload/upload.component.html
@@ -262,8 +262,9 @@
 <ng-template #cautionMessage>
     <div class="text-sm text-gray-500 font-normal">
         <strong class="text-sm text-gray-700 font-medium">CAUTION:</strong> Submitting new records will automatically
-        run attribute calculation and embeddings <span *ngIf="uploadSpecificHelper.hasGdpr">(recreation of the
-            embeddings will calculate them one more time)</span>
+        run attribute calculation and embeddings <span *ngIf="uploadSpecificHelper.hasNotGdprEmbeddings">(recreation of
+            the
+            embeddings will calculate them one more time and could cause additional fees)</span>
         for the newly uploaded records
     </div>
 </ng-template>
--- a/src/app/import/components/upload/upload.component.html
+++ b/src/app/import/components/upload/upload.component.html
@@ -262,7 +262,7 @@
 <ng-template #cautionMessage>
     <div class="text-sm text-gray-500 font-normal">
         <strong class="text-sm text-gray-700 font-medium">CAUTION:</strong> Submitting new records will automatically
-        run attribute calculation and embeddings <span *ngIf="uploadSpecificHelper.hasNotGdprEmbeddings">(recreation of
+        run attribute calculation and embeddings <span *ngIf="uploadSpecificHelper.recalculationCosts">(recreation of
             the
             embeddings will calculate them one more time and could cause additional fees)</span>
         for the newly uploaded records
--- a/src/app/import/components/upload/upload.component.html
+++ b/src/app/import/components/upload/upload.component.html
@@ -265,6 +265,6 @@
         run attribute calculation and embeddings <span *ngIf="uploadSpecificHelper.recalculationCosts">(recreation of
             the
             embeddings will calculate them one more time and could cause additional fees)</span>
-        for the newly uploaded records
+        for all records
     </div>
 </ng-template>
--- a/src/app/import/components/upload/upload.component.ts
+++ b/src/app/import/components/upload/upload.component.ts
@@ -1,6 +1,6 @@
 import { Component, ElementRef, EventEmitter, Input, OnChanges, OnDestroy, OnInit, Output, SimpleChanges, ViewChild } from '@angular/core';
 import { Router } from '@angular/router';
-import { timer } from 'rxjs';
+import { Subscription, timer } from 'rxjs';
 import { first } from 'rxjs/operators';
 import { PreparationStep } from 'src/app/base/components/upload-assistant/label-studio/label-studio-assistant-helper';
 import { LabelStudioAssistantComponent } from 'src/app/base/components/upload-assistant/label-studio/label-studio-assistant.component';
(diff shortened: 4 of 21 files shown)


src/app/projects/components/project-settings/components/embeddings/embeddings.component.html
```diff
@@ -166,12 +166,17 @@
...
+    disabled: settingModals.embedding.create.blocked || 
+    !(useableTextAttributes && settingModals.embedding.create.embeddingCreationFormGroup) || 
+    !settingModals.embedding.create.embeddingCreationFormGroup.get('termsAccepted').value ||
+    (settingModals.embedding.create.embeddingCreationFormGroup.get('platform').value == PlatformType.HUGGING_FACE && settingModals.embedding.create.embeddingCreationFormGroup.get('model').value == null) ||
+    (settingModals.embedding.create.embeddingCreationFormGroup.get('platform').value == PlatformType.OPEN_AI && (settingModals.embedding.create.embeddingCreationFormGroup.get('model').value == null || settingModals.embedding.create.embeddingCreationFormGroup.get('apiToken').value === null)) ||
+    (settingModals.embedding.create.embeddingCreationFormGroup.get('platform').value == PlatformType.COHERE && settingModals.embedding.create.embeddingCreationFormGroup.get('apiToken').value === null) ||
+    (settingModals.embedding.create.embeddingCreationFormGroup.get('platform').value == PlatformType.PYTHON && settingModals.embedding.create.embeddingCreationFormGroup.get('model').value == null),
```
JWittmeyer: this feels like way to much in the HTML file conditioning so it's not really readable. Can we create a disabled variable and use that & change it whenever necessary? 

I'm thinking of a function "isDisabled" that sets this condition to the disabled variable and is called on change of either of the values so it's more readable. Not linking a function in the HTML :)

src/app/projects/components/project-settings/components/embeddings/embeddings.component.ts
```diff
@@ -26,16 +30,28 @@ export class EmbeddingsComponent implements OnInit, OnDestroy, OnChanges {
...
+    return PlatformType;
+  }
+
+  constructor(private projectApolloService: ProjectApolloService, private organizationApolloService: OrganizationApolloService) { }
 
   ngOnInit(): void {
     this.prepareDownloadedModels();
+    this.organizationApolloService.getUserOrganization().pipe(first()).subscribe((organization) => {
+      this.organization = organization;
+    });
```
JWittmeyer, line 59: when this is called only on init a change wont be visible.
A function that is called here + on websocket notification

src/app/projects/components/project-settings/components/embeddings/embeddings.component.ts
```diff
@@ -65,18 +86,43 @@ export class EmbeddingsComponent implements OnInit, OnDestroy, OnChanges {
       }));
   }
 
+  prepareEmbeddingPlatforms() {
+    if (this.organization.gdprCompliant && this.isManaged) {
```
JWittmeyer: should only be organization.gdprCompliant since the managed state shouldn't effect the result (organizations on os should have the default correctly set)

src/app/projects/components/project-settings/components/embeddings/embeddings.component.ts
```diff
@@ -65,18 +86,43 @@ export class EmbeddingsComponent implements OnInit, OnDestroy, OnChanges {
...
+    if (platform == PlatformType.PYTHON) {
+      suggestionList = suggestionList.filter(e => e.configString == 'bag-of-words' || e.configString == 'bag-of-characters' || e.configString == 'tf-idf');
+      this.embeddingHandles[attId] = suggestionList;
+    } else if (platform == PlatformType.OPEN_AI) {
+      suggestionList = suggestionList.filter(e => e.configString == 'text-embedding-ada-002');
+      this.embeddingHandles[attId] = suggestionList;
+    } else if (platform == PlatformType.HUGGING_FACE) {
+      this.embeddingHandles[attId] = this.embeddingHandlesCopy[attId];
+      this.checkModelDownloaded();
+    }
```
JWittmeyer: these are filtered by a hardcoded config string. so if a new config string is added this result lin faulty values. That's why I mentioned that we should add a platformtype in the suggestions json from the backend so it's easier to filter :)

src/app/projects/components/project-settings/components/embeddings/embeddings.component.ts
```diff
@@ -65,18 +86,43 @@ export class EmbeddingsComponent implements OnInit, OnDestroy, OnChanges {
...
-        || (granularity == 'ON_TOKEN' && parseEl.token)) {
+      if ((granularity == EmbeddingType.ON_ATTRIBUTE && parseEl.attribute)
+        || (granularity == EmbeddingType.ON_TOKEN && parseEl.token)) {
         element.forceHidden = false;
       }
     }
+    this.selectedPlatform = this.embeddingPlatforms.find((p: EmbeddingPlatform) => p.platform == platform);
+    form.get('model').setValue(null);
+    form.get('apiToken').setValue(null);
+    form.get('termsAccepted').setValue(false);
```
JWittmeyer: not sure if i understand everything correctly. but when I already have a platform selected switching between model shouldn't result in clearing evything.

So not every change results the fields. Changing the platform, however, does.

src/app/projects/components/project-settings/components/embeddings/embeddings.component.ts
```diff
@@ -104,25 +150,42 @@ export class EmbeddingsComponent implements OnInit, OnDestroy, OnChanges {
...
+
+    const config = {
+      ...((platform == PlatformType.HUGGING_FACE || platform == PlatformType.PYTHON) && { model: embeddingForm.get("model").value }),
+      ...(platform == PlatformType.OPEN_AI && { model: embeddingForm.get("model").value, apiToken: embeddingForm.get("apiToken").value }),
+      ...(platform == PlatformType.COHERE && { apiToken: embeddingForm.get("apiToken").value }),
+      platform: platform,
+      termsText: this.gdprText.nativeElement.innerHTML,
+      termsAccepted: embeddingForm.get("termsAccepted").value,
+      embeddingType: embeddingForm.get("granularity").value.substring(3) === "TOKEN" ? EmbeddingType.ON_TOKEN : EmbeddingType.ON_ATTRIBUTE
+    };
```
JWittmeyer: imo not really readable because of the initial condition logic. 

```suggestion
    const config = {    
      platform: platform,
      termsText: this.gdprText.nativeElement.innerHTML,
      termsAccepted: embeddingForm.get("termsAccepted").value,
      embeddingType: embeddingForm.get("granularity").value.substring(3) === "TOKEN" ? EmbeddingType.ON_TOKEN : EmbeddingType.ON_ATTRIBUTE
    };
        
if (platform == PlatformType.HUGGING_FACE || ....){
   config.model = embeddingForm.get("model").value   
}   
...    
    
```

src/app/projects/components/project-settings/components/embeddings/embeddings.component.ts
```diff
@@ -104,25 +150,42 @@ export class EmbeddingsComponent implements OnInit, OnDestroy, OnChanges {
...
+
+    this.projectApolloService.createEmbedding(
+      this.project.id,
+      attributeId,
+      JSON.stringify(config)
+    ).pipe(first()).subscribe(() => {
+      this.settingModals.embedding.create.embeddingCreationFormGroup.get("platform").setValue(PlatformType.HUGGING_FACE);
+      this.settingModals.embedding.create.embeddingCreationFormGroup.get("model").setValue(null);
+      this.settingModals.embedding.create.embeddingCreationFormGroup.get("apiToken").setValue(null);
```
JWittmeyer: i think this should be a function "initEmbeddingModal" or something like that with e.g. two options (full init = everything clear, platform switch = API token, ...  is cleared)

src/app/projects/components/project-settings/components/embeddings/embeddings.component.ts
```diff
@@ -104,25 +150,42 @@ export class EmbeddingsComponent implements OnInit, OnDestroy, OnChanges {
...
+    });
   }
 
   selectFirstUnhiddenEmbeddingHandle(inputElement: HTMLInputElement) {
-    const suggestionList = this.embeddingHandles[this.settingModals.embedding.create.embeddingCreationFormGroup.get("targetAttribute").value];
-    for (let embeddingHandle of suggestionList) {
-      if (!embeddingHandle.hidden && !embeddingHandle.forceHidden) {
-        this.selectEmbeddingHandle(embeddingHandle, inputElement);
+    let suggestionList = this.embeddingHandles[this.settingModals.embedding.create.embeddingCreationFormGroup.get("targetAttribute").value];
```
JWittmeyer: is there a specific reason this was changed to a let?

src/app/projects/components/project-settings/components/embeddings/embeddings.component.ts
```diff
@@ -132,22 +195,22 @@ export class EmbeddingsComponent implements OnInit, OnDestroy, OnChanges {
 
   checkEmbeddingHandles(eventTarget: HTMLInputElement) {
     const embeddingForm = this.settingModals.embedding.create.embeddingCreationFormGroup;
-    embeddingForm.get('embeddingHandle').setValue(eventTarget.value);
-    const suggestionList = this.embeddingHandles[embeddingForm.get("targetAttribute").value];
+    embeddingForm.get('model').setValue(eventTarget.value);
+    let suggestionList = this.embeddingHandles[embeddingForm.get("targetAttribute").value];
```
JWittmeyer: same here?

src/app/projects/components/project-settings/components/gates-integrator/gates-integrator.component.html
```diff
@@ -42,4 +42,17 @@
...
+
+<kern-modal [isOpen]="gatesIntegratorModals.warningModal.open" closeButton="X"
+    [modalBoxStyle]="{'width':'30rem','text-align':'center'}" [acceptButton]="{
+    buttonCaption: 'Accept',
+    emitObject: this, emitFunction: updateProjectForGates
+}" (optionClicked)="gatesIntegratorModals.warningModal.open = false">
+    <h1 class="text-lg text-gray-900 mb-2">Confirmation</h1>
+    <div class="text-sm text-gray-500 my-2">
+        Are you sure you want to run the gates integration?
+        <p>This will cause double fees since embeddings will be recreated.</p>
```
JWittmeyer: double fees

src/app/projects/components/project-settings/helper/data-handler-helper.ts
```diff
@@ -84,8 +87,14 @@ export class DataHandlerHelper {
         const values = settingModals.embedding.create.embeddingCreationFormGroup.getRawValue();
         let toReturn = this.getAttributeArrayAttribute(values.targetAttribute, 'name', attributes);
         toReturn += "-" + (values.granularity == 'ON_ATTRIBUTE' ? 'classification' : 'extraction');
-        toReturn += "-" + values.embeddingHandle;
-
+        const platform = settingModals.embedding.create.embeddingCreationFormGroup.get('platform').value
+        if (platform == PlatformType.HUGGING_FACE || platform == PlatformType.PYTHON) {
+            toReturn += "-" + values.model;
+        } else if (platform == PlatformType.OPEN_AI) {
+            toReturn += "-" + values.model + "-" + values.apiToken;
```
JWittmeyer: i think the api token isn't (or at least shouldn't) be used in the final name. Not sure what the logic you and simon decided on was for the name but I think some kind of obfuscation or reduced values or something similar. Since names need to be unique this should be included here as well.

src/app/projects/components/project-settings/helper/data-handler-helper.ts
```diff
@@ -84,8 +87,14 @@ export class DataHandlerHelper {
...
+        } else if (platform == PlatformType.COHERE) {
+            toReturn += "-" + values.apiToken;
```
JWittmeyer: see above

src/app/projects/components/project-settings/helper/project-settings-helper.ts
```diff
@@ -1,10 +1,29 @@
...
+    { name: 'Attribute', value: EmbeddingType.ON_ATTRIBUTE },
+    { name: 'Token', value: EmbeddingType.ON_TOKEN }
 ];
 
+export enum PlatformType {
+    HUGGING_FACE = "huggingface",
+    OPEN_AI = "openai",
+    COHERE = "cohere",
+    PYTHON = "python"
+}
```
JWittmeyer, line 19: if we are using a string enum with different values from our standard (e.g. HUGGING_FACE = "HUGGING_FACE") I need to understand why. (I'm assuming the backend has it like this but don't know yet :D ). 

-> A comment would help here

JWittmeyer:
not sure why but for submodule the tree not the pr is linked

- [x] resovled

JWittmeyer:
wouldn't the base branch be new-embedding-providers instead of dev?

- [ ] resolved

src/app/projects/components/project-settings/components/embeddings/embeddings.component.ts
lumburovskalina: Yes switching between models does not result in clearing everything. Currently it clears the selected options when switching between platform and granularity. Do we want this for granularity?

JWittmeyer:
I think we already talked about these points earlier this week but (because of my faux pas this morning) these aren't yet included. So for completion sake since I noticed it during testing:
1. Token Granulatiry doesn't work with all platforms
2. not all platforms have a tos thing
3. models for huggingface aren't filtered correctly

image


- [x] resolved

JWittmeyer:
I would default to platform huggingface since that's the one we used so far most of the time

- [x] resolved

src/app/projects/components/project-settings/components/embeddings/embeddings.component.ts
JWittmeyer: misunderstanding. the method is call checkForceHiddenHandles. i thought this was called on keydown of the combination boxes (dropdown with search). Thats the hidden flag, not the forcehidden one. Sorry :D can be resolved

src/app/projects/components/project-settings/components/embeddings/embeddings.component.ts
lumburovskalina: no reason, probably because of some testing but I didn't return it back, changed it to const

src/app/projects/components/project-settings/components/embeddings/embeddings.component.ts
lumburovskalina: changed it to const

src/app/base/services/project/project-apollo.service.ts
```diff
@@ -1489,4 +1490,17 @@ export class ProjectApolloService {
         })
       );
   }
+
+  getEmbeddingPlatforms(): Observable<EmbeddingPlatform> {
+    return this.apollo
+      .query({
+        query: queries.GET_EMBEDDING_PLATFORMS,
+        fetchPolicy: 'no-cache',
```
JWittmeyer: i think the platforms wont change within a build so no need to always rerequest them. Maybe 'cache-first' ?

src/app/import/components/helpers/upload-specific.ts
```diff
@@ -26,6 +26,7 @@ export class RecordNewUploadHelper {
 
 export class RecordAddUploadHelper {
     projectName: string = '';
+    hasNotGdprEmbeddings: boolean = false;
```
JWittmeyer: double negative is hard to read so maybe a flip can help 
```suggestion
    hasGdprEmbeddings: boolean = true;
```

src/app/import/components/upload/upload.component.html
```diff
@@ -262,7 +262,9 @@ <h3 class="text-sm font-medium text-yellow-800">File required</h3>
 <ng-template #cautionMessage>
     <div class="text-sm text-gray-500 font-normal">
         <strong class="text-sm text-gray-700 font-medium">CAUTION:</strong> Submitting new records will automatically
-        run attribute calculation and embeddings for the newly
-        uploaded records
+        run attribute calculation and embeddings <span *ngIf="uploadSpecificHelper.hasNotGdprEmbeddings">(recreation of
```
JWittmeyer: is this correct? The gdpr flag decided if it costs more? Shouldn't this be depending on the model provider (cohere & openai?)

src/app/import/components/upload/upload.component.ts
```diff
@@ -300,4 +303,13 @@ export class UploadComponent implements OnInit, OnChanges, OnDestroy {
     this.uploadSpecificHelper.selectedTokenizer = findName.name;
   }
 
+  prepareEmbeddings(): void {
+    let q, vc;
+    [q, vc] = this.projectApolloService.getEmbeddingSchema(this.projectId);
+    vc.pipe(first()).subscribe((embeddings: Embedding[]) => {
+      const hasNotGdprEmbeddings = embeddings.filter((e: Embedding) => e.platform == PlatformType.COHERE || e.platform == PlatformType.OPEN_AI).length > 0;
+      this.uploadSpecificHelper.hasNotGdprEmbeddings = hasNotGdprEmbeddings;
```
JWittmeyer: ah it's just the name of the variable that might be misleading :D functionality is already like described above.

Maybe name "recalculationCosts" or something similar?

src/app/import/components/upload/upload.component.ts
```diff
@@ -300,4 +303,13 @@ export class UploadComponent implements OnInit, OnChanges, OnDestroy {
     this.uploadSpecificHelper.selectedTokenizer = findName.name;
   }
 
+  prepareEmbeddings(): void {
+    let q, vc;
+    [q, vc] = this.projectApolloService.getEmbeddingSchema(this.projectId);
+    vc.pipe(first()).subscribe((embeddings: Embedding[]) => {
+      const hasNotGdprEmbeddings = embeddings.filter((e: Embedding) => e.platform == PlatformType.COHERE || e.platform == PlatformType.OPEN_AI).length > 0;
```
JWittmeyer: since we don't actually need the whole list (just want to know if at least one exists) i'd go with `some` instead of a full filter (which in turn creates a new list etc.)

src/app/projects/components/project-settings/components/embeddings/embeddings.component.ts
```diff
@@ -104,25 +154,50 @@ export class EmbeddingsComponent implements OnInit, OnDestroy, OnChanges {
   addEmbedding() {
     if (!this.dataHandlerHelper.canCreateEmbedding(this.settingModals, this.embeddings, this.attributes)) return;
     const embeddingForm = this.settingModals.embedding.create.embeddingCreationFormGroup;
-    const embeddingHandle = embeddingForm.get("embeddingHandle").value;
     const attributeId = embeddingForm.get("targetAttribute").value;
-    const granularity = embeddingForm.get("granularity").value;
+    const platform = embeddingForm.get("platform").value;
+
+    if (this.gdprText) {
+      this.gdprText.nativeElement.innerHTML = this.gdprText.nativeElement.innerHTML.split(':')[0] + ":" + this.anchorGdpr.nativeElement.innerHTML;
```
JWittmeyer: changing an innerHTML isn't recommended for text changes. Why not create a variable gdprTextHTML or something like that and fill it here + in HTML file add an *if for if the variable is filled?

src/app/projects/components/project-settings/helper/data-handler-helper.ts
```diff
@@ -84,20 +87,40 @@ export class DataHandlerHelper {
...
+            } else if (values.apiToken.length > 8) {
+                toReturn += "-" + values.model + "-" + values.apiToken?.substring(0, 8) + "...";
+            } else {
+                toReturn += "-" + platform + "-" + values.model + "-" + values.apiToken?.substring(0, values.apiToken.length);
+            }
+        } else if (platform == PlatformType.COHERE) {
+            if (values.apiToken == null || values.apiToken.length == 0) {
+                toReturn += "-" + platform + "-" + values.apiToken;
+            } else if (values.apiToken.length > 8) {
+                toReturn += "-" + platform + "-" + values.apiToken?.substring(0, 8) + "...";
```
JWittmeyer: outdated logic for the api key.

Also this seems like a lot of similar code between the cases. 

Also I don't think we need to check for the length since substring of something with less length will result in the remaining string anyway.

Also also. why do we need to check `values.apiToken?` if we already know it exists (otherwise the length check before wouldn't work)

src/app/projects/components/project-settings/project-settings.component.ts
```diff
@@ -319,11 +324,15 @@ export class ProjectSettingsComponent implements OnInit, OnDestroy {
     this.projectApolloService.getProjectSize(projectId).pipe(first()).subscribe((size) => {
       this.settingModals.projectExport.projectSize = size;
       size.forEach((element) => {
+        let hasGdpr = false;
+        if (element.table == 'embedding tensors') {
+          hasGdpr = this.embeddings.filter((e: any) => e.name.split("-")[2] == PlatformType.COHERE || e.name.split("-")[2] == PlatformType.OPEN_AI).length > 0;
```
JWittmeyer: some function should do the trick

JWittmeyer:
I think we need to reorder provider and granulatiriy in the modal since provider decides what ganularity is possible

- [x] resolved

JWittmeyer:
As far as i know we discussed that the end user should get some hints on what to look out for in the terms of service. Like "look for section 4 a" for cohere (or something like that). Where the texts discussed with Henrik?

- [x] resolved

Lina: Yes, the texts were discussed with Henrik --> will be rediscussed 
Lina: I removed the texts on the front-end, and added them as discussed before,  I will retest this once the messages are rediscussed

JWittmeyer:
The check if i am allowed to add an embedding doesn't seem to work. (most likely because of the slightly changed logic in name generation so if [this](https://github.com/code-kern-ai/refinery-ui/pull/143#discussion_r1229304730) is solved this should be solved too)

image

- [x] resolve

JWittmeyer:
Changing a target attribute cleared the model & API token.

I think for the attribute change we shouldn't clear since it's frustrating to reenter everything if i only want a different target

- [x] resolved

JWittmeyer:
image

this is not just for the newly uploaded records but for all records

- [x] resolved

src/app/projects/components/project-settings/components/embeddings/embeddings.component.ts
lumburovskalina: Yes we can add the variable here so we would avoid changing the innerHTML.
I think the check in the template is not needed since there we have a check if we want to display whole 'terms and conditions' part and here we check if that is displayed we take the message from there.

Maybe we could add the messages as constants and display them in the template and use them here instead of the whole implementation with the innerHTML text.

JWittmeyer: i think the text were specifically implemented to be transferred from backend so having a constant in the UI wouldn't make sense. However I'm not sure if that's what you meant

lumburovskalina: Okay yes, that's what I meant. Since it is transferred from the BE, the innerHTML and the constants will be removed.

src/app/projects/components/project-settings/helper/data-handler-helper.ts
```diff
@@ -84,20 +87,34 @@ export class DataHandlerHelper {
...
+            toReturn += this.buildEmbeddingNameWithApiToken(values, platform);
+        }
+        return toReturn;
+    }
 
+    buildEmbeddingNameWithApiToken(values: any, platform: string) {
+        if (values.apiToken == null) return "";
+        let toReturn = "";
+        const apiTokenCut = values.apiToken.substring(0, 3) + "..." + values.apiToken.substring(values.apiToken.length - 4, values.apiToken.length);
+        toReturn += `-${platform}-${platform == PlatformType.OPEN_AI ? values.model + "-" + apiTokenCut : apiTokenCut}`;
```
JWittmeyer: += not necessary if nothing exists before

Also I don't like iterpolation strings with conditions inside since they are hard to read.

Maybe something like:

```suggestion
        if (values.apiToken == null) return "";
        const platformStr = "-" + platform + "-";
        const apiTokenCut = values.apiToken.substring(0, 3) + "..." + values.apiToken.substring(values.apiToken.length - 4, values.apiToken.length);
        if (platform == PlatformType.OPEN_AI) return platformStr + values.model + "-" + apiTokenCut;
        else return platformStr + apiTokenCut;
```

JWittmeyer (approved):

PR #144, opened 2023-06-13
Linebreaks in labeling suite

### LINKED ISSUE #256: [BUG] - Linebreaks are not displayed correctly in the labeling view
**Describe the bug**
I have a pandas dataframe that has an attribute "email" which are full-length E-Mails. These E-Mails have a typical structure with a greeting, a body, an ending and a footer. The nice formatting is achieved with "\n" characters in my Python strings. When I export my data to csv (quoting=1) or to json using the pandas functions and then upload it to refinery, the formatting gets partially lost. The displayed E-Mails are now just a single blob without any linebreaks whatsoever.

After testing with a minimal example, I found that "\n" is displayed correctly, but "\n\n" does not display ANY linebreak at all.

**To Reproduce**
1. Go to refinery
2. Upload my sample data as a new project
3. Create Extraction labeling task
4. Go to labeling view and see error

**Expected behavior**
All linebreaks should be displayed correctly, no matter how many are chained. 

**Screenshots**
image

 - OS:  win
 - Browser edge
 - Version: dev-app v1.10.0

**EXAMPLE.json**
```
[{"Summary":"Hello there,\n\nGeneral Kenobi\n\nFOOTER\nFooter2\nFooter3\nFooter4\n\nFooter5"}]
```

lumburovskalina:
- Closes: https://github.com/code-kern-ai/refinery/issues/256 

PR:
- https://github.com/code-kern-ai/refinery-ui/pull/144
- https://github.com/code-kern-ai/submodule-javascript-functions/pull/1

### DIFF
--- a/src/app/base/components/record-display/record-display.component.html
+++ b/src/app/base/components/record-display/record-display.component.html
@@ -13,6 +13,7 @@
         </ng-template>
         <ng-template #noHighlight>
             <span
+                [ngClass]="recordDisplayOptions.dataBrowserModals.configuration.lineBreaks.normal ? recordDisplayOptions.dataBrowserModals.configuration.lineBreaks.isPreWrap? 'whitespace-pre-wrap':'whitespace-pre-line' :''"
                 *ngIf="record.data[attributes[attribute.key].name] != null && record.data[attributes[attribute.key].name] !== ''; else notPresent">
                 {{record.data[attributes[attribute.key].name]}}
             </span>
--- a/src/app/base/components/record-display/record-display.component.html
+++ b/src/app/base/components/record-display/record-display.component.html
@@ -13,7 +13,8 @@
         </ng-template>
         <ng-template #noHighlight>
             <span
-                [ngClass]="recordDisplayOptions.dataBrowserModals.configuration.lineBreaks.normal ? recordDisplayOptions.dataBrowserModals.configuration.lineBreaks.isPreWrap? 'whitespace-pre-wrap':'whitespace-pre-line' :''"
+                [ngClass]="recordDisplayOptions.dataBrowserModals.configuration.lineBreaks != LineBreaksType.NORMAL ? 
+                (recordDisplayOptions.dataBrowserModals.configuration.lineBreaks == LineBreaksType.IS_PRE_WRAP ? 'whitespace-pre-wrap' : 'whitespace-pre-line') : '' "
                 *ngIf="record.data[attributes[attribute.key].name] != null && record.data[attributes[attribute.key].name] !== ''; else notPresent">
                 {{record.data[attributes[attribute.key].name]}}
             </span>
--- a/src/app/base/components/record-display/record-display.component.html
+++ b/src/app/base/components/record-display/record-display.component.html
@@ -16,7 +16,7 @@
                 [ngClass]="recordDisplayOptions.dataBrowserModals.configuration.lineBreaks != LineBreaksType.NORMAL ? 
                 (recordDisplayOptions.dataBrowserModals.configuration.lineBreaks == LineBreaksType.IS_PRE_WRAP ? 'whitespace-pre-wrap' : 'whitespace-pre-line') : '' "
                 *ngIf="record.data[attributes[attribute.key].name] != null && record.data[attributes[attribute.key].name] !== ''; else notPresent">
-                {{record.data[attributes[attribute.key].name]}}
+                <span>{{record.data[attributes[attribute.key].name]}}</span>
             </span>
             <ng-template #notPresent>
                 <svg xmlns="http://www.w3.org/2000/svg"
--- a/src/app/base/services/record/record-apollo.service.ts
+++ b/src/app/base/services/record/record-apollo.service.ts
@@ -533,7 +533,7 @@ export class RecordApolloService {
   }
 
   //private
-  #tokenMapper(token, tokensLength) {
+  #tokenMapper(token: any, tokensLength: number) {
     let countLineBreaks = 0;
     if (token.value.includes("\n")) {
       countLineBreaks = token.value.split("\n").length - 1;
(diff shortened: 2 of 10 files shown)


src/app/base/components/record-display/record-display.component.html
```diff
@@ -13,6 +13,7 @@
         </ng-template>
         <ng-template #noHighlight>
             <span
+                [ngClass]="recordDisplayOptions.dataBrowserModals.configuration.lineBreaks.normal ? recordDisplayOptions.dataBrowserModals.configuration.lineBreaks.isPreWrap? 'whitespace-pre-wrap':'whitespace-pre-line' :''"
```
JWittmeyer: condition looks a bit weird to me. I read it as if its normal we then decide between pre wrap & pre line (since ? is <condition> ? <true> : <false>. maybe some brackets or !normal

src/app/base/services/record/record-apollo.service.ts
```diff
@@ -533,13 +533,26 @@ export class RecordApolloService {
   }
 
   //private
-  #tokenMapper(token) {
+  #tokenMapper(token: any, tokensLength: number) {
+    let countLineBreaks = 0;
+    if (token.value.includes("\n")) {
+      countLineBreaks = token.value.split("\n").length - 1;
```
JWittmeyer: splitting of a string costs since an array with new strings is created.
since we are doing this per token this can be a bit much. 

fasten way I know of would be a creating a small function "countOccurrance" & looping over the string
```
function countOcc(str, search){
    let c = 0,p=-1;
    while (true){
        p = str.indexOf(search,p+1)
        if(p!= -1) c++;
        else break;
    }
    return c;
}
```

src/app/base/services/record/record-apollo.service.ts
```diff
@@ -533,13 +533,26 @@ export class RecordApolloService {
...
+      countLineBreaks = checkIfOrLastIdx ? countLineBreaks : countLineBreaks - 1;
+    }
     return {
       value: token.value,
       idx: token.idx,
       posStart: token.posStart,
       posEnd: token.posEnd,
       type: token.type,
+      countLineBreaks: countLineBreaks,
+      countLineBreaksArray: countLineBreaks != 0 ? Array(countLineBreaks - 1).fill(0) : null,
```
JWittmeyer: why do we need this?
looks to me like ist an array with 0 (or null)
we can also easily just loop between 0 and countLineBreak in most cases. Since it's for the HTML page we need an array of items (I think) :D 

==> add comment

src/app/data/components/data-browser/helper-classes/modals-helper.ts
```diff
@@ -32,6 +32,10 @@ export type ConfigurationModal = {
     highlightText: boolean;
     weakSupervisionRelated: boolean;
     separator: string;
+    lineBreaks: {
+        normal: boolean;
+        isPreWrap: boolean;
+    };
```
JWittmeyer: i think this would make more sense as an enum since we can'T have normal and isPreWrapped = true

src/app/base/services/record/record-apollo.service.ts
lumburovskalina: should we add this as part of the submodules? :D maybe we would need it for something else too

JWittmeyer: yes, sounds good :)

src/app/base/services/record/record-apollo.service.ts
lumburovskalina: There are two options how we can do this: here in the .ts file with creating an array from the number of elements or we can just send the number of iterations here and use [].constructor(countLineBreaks) as an array to loop.
I think keeping the html more clean is a better way.

I will add a comment here, the fill(0) part is not needed yes, only the creation of the array

lumburovskalina: Comment added and fill(0) removed :)

src/app/base/services/record/record-apollo.service.ts
```diff
@@ -533,13 +534,23 @@ export class RecordApolloService {
...
 
   //private
-  #tokenMapper(token) {
+  #tokenMapper(token: any, tokensLength: number) {
+    let countLineBreaks = countOcc(token.value, "\n");
+    if (countLineBreaks > 0) {
+      // If we are on the first or last token and either/both of them is new lines, the class w-full cannot work because we don't have a previous or next token, that's why we need the original countLineBreaks
+      // If we are not on the first or last token, the array of the countLineBreaks has to be one less than the actual countLineBreaks because we use the class w-full of the current line as one line break
+      // Adding a completely new line and having a text that needs a new line are different in terms of css classes
+      const checkIfOrLastIdx = token.idx == 0 || token.idx == tokensLength;
```
JWittmeyer: might be wrong but isn't token length to much?
idx starts at 0 so if I understand it correctly it should never be == tokenLength but == tokenLength -1

JWittmeyer:
To automatically close the Issue we can use implements in the [same line as the link](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)

- [x] resolved

JWittmeyer:
Since display pre interpretes whitespace as "this needs to be displayed" we now have the issue that spaces are shown even though no space is present in the record
image


I think we need to put the text without any whitespace
image

so something like `<span>{somevalue}</span>`

- [x] resolved

src/app/base/services/record/record-apollo.service.ts
lumburovskalina: yes that is true, i added this as a lastIdx param so we are not confused and extract the minus -1 when calling the func

JWittmeyer: alright :D

JWittmeyer (approved):

PR #145, opened 2023-06-19
Sub-component for crypted field

lumburovskalina:
PRs:
- https://github.com/code-kern-ai/refinery-gateway/pull/137
- https://github.com/code-kern-ai/refinery-submodule-model/pull/45
- https://github.com/code-kern-ai/refinery-ui/pull/145
- https://github.com/code-kern-ai/dev-setup/pull/30


src/app/base/components/crypted-field/crypted-field.component.html
```diff
@@ -0,0 +1,36 @@
+<div class="form-control">
+    <label class="text-gray-500 text-sm font-medium m-1">
+        Use crypted key for this project <em>- optional</em>
```
JWittmeyer: frontend text can be password, just didn't want the field in the db to be called "password"

JWittmeyer: also why is it "for this project"? I would expect Something like "Password:" as an optional input parameter for the component that can be overwrite it if necessary.

src/app/base/components/export/export.component.ts
```diff
@@ -353,7 +354,7 @@ export class ExportComponent implements OnInit, OnChanges {
     const jsonString = this.exportHelper.buildExportData();
     if (this.exportHelper.error.length != 0) return
     this.downloadState = DownloadState.PREPARATION;
-    this.projectApolloService.prepareRecordExport(this.projectId, jsonString).pipe(first()).subscribe((x) => {
+    this.projectApolloService.prepareRecordExport(this.projectId, jsonString, this.key).pipe(first()).subscribe((x) => {
```
JWittmeyer: to prevent issues i think we should check for the key, if it's an empty string we send "null" so it's a bit more clear that it's not a password protected file

src/app/import/components/upload/upload.component.html
```diff
@@ -63,6 +64,7 @@
     <ng-container *ngTemplateOutlet="dataSourcesUpload"></ng-container>
     <ng-container *ngTemplateOutlet="upload"></ng-container>
     <ng-container *ngTemplateOutlet="fileValidation"></ng-container>
+    <ng-container *ngTemplateOutlet="cryptedField"></ng-container>
```
JWittmeyer: since the kern-crypt field is a one liner anyway why is this put into a container?

src/app/projects/components/project-settings/project-settings.component.html
```diff
@@ -214,18 +214,15 @@
             </div>
         </ng-template>
     </div>
-    <div *ngIf="settingModals.projectExport.projectSize" class="mt-4 flex flex-row justify-between "
+    <div *ngIf="settingModals.projectExport.projectSize" class="mt-4 justify-between"
```
JWittmeyer: i think jusitfy-between needs a flex or something like that so we can maybe remove the class?

JWittmeyer:
Not sure if i want an auto-complete based on my login passwords. 
Personally, it would irritate me. Especially since no one should use their kern login password for a "one off" zip file

image


- [x] resolved

JWittmeyer:
Maybe "Encrypt zip file with password - optional" ?

image

- [x] resolved

JWittmeyer:
if it's a plain json file we can't use a password. I think we should hide the input depending on the file type (zip yes, rest no). Same for the normal upload

image

- [x] resolved

src/app/base/components/crypted-field/crypted-field.component.html
lumburovskalina: Okay yes the default label will be Password, and we use that on the import and export of the records, but for project export it will be: "Encrypt zip file with password - optional" right?

JWittmeyer: yes

src/app/import/components/upload/upload.component.html
lumburovskalina: no need for container, I changed this to template

src/app/projects/components/project-settings/project-settings.component.html
lumburovskalina: yes, that's true, the class does not have any impact on the element since it does not have the flex display, removed :D

src/app/base/components/export/export.component.ts
```diff
@@ -353,7 +354,8 @@ export class ExportComponent implements OnInit, OnChanges {
     const jsonString = this.exportHelper.buildExportData();
     if (this.exportHelper.error.length != 0) return
     this.downloadState = DownloadState.PREPARATION;
-    this.projectApolloService.prepareRecordExport(this.projectId, jsonString).pipe(first()).subscribe((x) => {
+    if (this.key == '') this.key = null;
```
JWittmeyer: with this we change the class instance of key. What I meant was 

```suggestion
    let keyToSend = this.key;
    if (!keyToSend) = keyToSend = null;
```
or some other variation of this behaviour to ensure that the class instance isn't changed but the values is "skipped" when send to backend

JWittmeyer:
Password is still autocompleted. And with this fills my login password
image

 If I can believe [this](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion) link modern browser ignore the autocomplete off property because of some reasons and try to fill it anyway 😁  

[further down](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#preventing_autofilling_with_autocompletenew-password) they mention that new-password might do the trick. 

If it doesn't work we could have two text fields, if the password isn't visible we use a different string than the password string that is prepared alongside the actual password string and holds only the char ●.

- [x] resolved

JWittmeyer:
Can't enter passwords on upload anymore
Screenshot from snapshot upload
image

(same on normal file upload)

I'm assuming that the type isn't correct (might be a windows) Maybe we can change this condition to file && (type zip || file name ends with .zip)
image



- [x] resolved

src/app/import/components/upload/upload.component.html
```diff
@@ -267,4 +269,9 @@ <h3 class="text-sm font-medium text-yellow-800">File required</h3>
             embeddings will calculate them one more time and could cause additional fees)</span>
         for all records
     </div>
+</ng-template>
+
+<ng-template #cryptedField>
+    <kern-crypted-field *ngIf="file && file.type == (zipType || fileEndsWithZip)"
```
JWittmeyer: doesn't work like this. the brackets are wrong

JWittmeyer:
Encrypt zip file with password - optional
should also be present for export of records (databrowser & settings)

- [x] resolved

JWittmeyer:
Password solution doesn't work
rec2

- [x] resolved

src/app/base/components/crypted-field/crypted-field.component.ts
```diff
@@ -0,0 +1,60 @@
...
+  toggleKey() {
+    this.show = !this.show;
+  }
+
+
+  onBackspace(event) {
+    if (event.key === 'ArrowLeft' || event.key === 'ArrowRight') {
+      return;
+    }
+    if (this.inputElement.nativeElement.selectionStart === 0 && !(event.ctrlKey || event.metaKey)) {
```
JWittmeyer: this has some issues with partial selection etc. 
I don't think there is an easy solution (a lot of discussion on stack overflow though :D ) nothing really worked when I tested. The best solution I found was with a field and the "password" field is a text field that is changed to password on click -> this will still sometimes result in the display of the password manager & question to update a password but that's better than using a different password than I think I'm using.

A different approach could be to use a **** font but that has another slew of security issues.

I checked in the changes and will approve the pr

JWittmeyer (approved):

PR #146, opened 2023-06-22
Remove GDPR flag on bricks integrator

lumburovskalina:
PRs:
- https://github.com/code-kern-ai/refinery-ui/pull/146/
- https://github.com/code-kern-ai/submodule-javascript-functions/pull/2

NOTE:
The active learners currently do not work on the heuristics page, will be fixed as part of bricks: https://github.com/code-kern-ai/bricks/pull/320/ 
NOTE:
https://github.com/code-kern-ai/refinery-ui/pull/146#issuecomment-1610912440 -> will be fixed also as part of bricks

### DIFF
--- a/src/app/base/components/bricks-integrator/bricks-integrator.component.ts
+++ b/src/app/base/components/bricks-integrator/bricks-integrator.component.ts
@@ -13,7 +13,7 @@ import { BricksAPIData, BricksIntegratorConfig, BricksSearchData, BricksVariable
 import { REMOVE_GROUP_NODES, extendDummyElements, getDummyNodeByIdForApi } from './helper/dummy-nodes';
 import { caesarCipher } from 'src/app/util/cipher';
 import { PASS_ME } from 'src/app/util/cipher';
-import { copyToClipboard, filterArrayMinusExcluded, isStringTrue, jsonCopy } from 'submodules/javascript-functions/general';
+import { copyToClipboard, removeArrayFromArray, isStringTrue, jsonCopy } from 'submodules/javascript-functions/general';
 import { capitalizeFirst } from 'submodules/javascript-functions/case-types-parser';
 
 @Component({
@@ -143,10 +143,15 @@ export class BricksIntegratorComponent implements OnInit, OnDestroy {
     const splitVal: string[] = value.split(',');
     for (let i = 0; i < splitVal.length; i++) {
       filter += "&filters[" + attribute + "][$eq]=" + splitVal[i].trim();
-      // Remove active learners from generators (on ac page we have generators and classifiers but we want to exclude active learners there)
-      if (splitVal[i].trim() == 'generator' && this.executionTypeFilter != "activeLearner") {
-        filter += "&filters[executionType][$ne]=activeLearner";
-      }
+      filter += this.filterActiveLearnersFromGenerators(splitVal, i, filter);
+    }
+    return filter;
+  }
+
+  private filterActiveLearnersFromGenerators(splitVal: string[], index: number, filter: string) {
+    // Remove active learners from generators (on ac page we have generators and classifiers but we want to exclude active learners there)
+    if (splitVal[index].trim() == 'generator' && this.executionTypeFilter != "activeLearner") {
+      filter += "&filters[executionType][$ne]=activeLearner";
     }
     return filter;
   }
@@ -174,15 +179,15 @@ export class BricksIntegratorComponent implements OnInit, OnDestroy {
         this.config.search.currentRequest = null;
         let finalData;
         if (this.config.extendedIntegrator) {
-          finalData = data.data.map(e => {
-            const c = jsonCopy(e);
-            if (e.attributes.partOfGroup) {
-              c.attributes.partOfGroup = JSON.parse(c.attributes.partOfGroup);
-              c.attributes.partOfGroup = filterArrayMinusExcluded(c.attributes.partOfGroup, REMOVE_GROUP_NODES);
+          finalData = data.data.map(integratorData => {
+            const integratorDataCopy = jsonCopy(integratorData);
+            if (integratorData.attributes.partOfGroup) {
+              integratorDataCopy.attributes.partOfGroup = JSON.parse(integratorDataCopy.attributes.partOfGroup);
+              integratorDataCopy.attributes.partOfGroup = removeArrayFromArray(integratorDataCopy.attributes.partOfGroup, REMOVE_GROUP_NODES);
             }
-            if (e.attributes.availableFor) c.attributes.availableFor = JSON.parse(c.attributes.availableFor);
-            if (e.attributes.integratorInputs) c.attributes.integratorInputs = JSON.parse(c.attributes.integratorInputs);
-            return c;
+            if (integratorData.attributes.availableFor) integratorDataCopy.attributes.availableFor = JSON.parse(integratorDataCopy.attributes.availableFor);
+            if (integratorData.attributes.integratorInputs) integratorDataCopy.attributes.integratorInputs = JSON.parse(integratorDataCopy.attributes.integratorInputs);
+            return integratorDataCopy;
           });
         } else finalData = data.data;
         if (this.executionTypeFilter) {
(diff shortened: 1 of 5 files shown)


SimonDegrafKern:
Not really GDPR-related, but for me when I stepped into this overview it was striking that the "details" just look like a field and not a header of a group of fields and that there is no space between the parantheses and "source code"
Bildschirmfoto 2023-06-27 um 11 17 15

- [x] resolved

Lina: For the source code empty space I would agree :)
For the Details header, I would assume that the cursor pointer on "Details" shows that it's clickable and the arrow on the right side, but still maybe we could underline or bold it 
I think we can make both 'Details' and 'Source code' underline 


Simon:
I see where you you are coming from but for me it just looks like the empty space is missing something. Maybe there can be something like (collapsed) or (click to open) or something like this be put as a placeholder? Or the underlined solution you proposed 
image

SimonDegrafKern:
Again not GDPR related but source code seems to be cut off and I could not (or did not manage to) scroll

Bildschirmfoto 2023-06-27 um 11 19 24

- [x] resolved

submodules/javascript-functions/general.ts
```diff
@@ -173,4 +173,8 @@ export function countOccurrences(str: string, search: string): number {
         else break;
     }
     return c;
-} 
\ No newline at end of file
+}
+
+export function filterArrayMinusExcluded(array: any[], excludedArray: any[]) {
```
SimonDegrafKern: Maybe just personal preference but for me this name was confusing because I thought it will filter values out of an array and skips some excluded ones from filtering. I would have used a signature e.g.

`export function filterArray(array: any[], filterValues: any[]) {`

Therefore it would be clear for me that I will filter the values out of the array. But just as an idea, you can of course decide on what you prefer.

src/app/base/components/bricks-integrator/helper/dummy-nodes.ts
```diff
@@ -215,4 +215,7 @@ def vader_sentiment(record):
         elif vs["compound"] <= -0.05:
             return "negative"
     elif MODE == "scores": 
-        return vs`;
\ No newline at end of file
+        return vs`;
+
+
+export const REMOVE_GROUP_NODES = ['gdpr_compliant', 'not_gdpr_compliant'];
```
SimonDegrafKern: Will this be removed in the future when the whole GDPR topic is removed from bricks? If yes, maybe add a # TODO comment. If we forget about this we will have these terms, which we want to remove, in the source code for a long time

src/app/base/components/bricks-integrator/bricks-integrator.component.ts
```diff
@@ -143,6 +143,10 @@ export class BricksIntegratorComponent implements OnInit, OnDestroy {
     const splitVal: string[] = value.split(',');
     for (let i = 0; i < splitVal.length; i++) {
       filter += "&filters[" + attribute + "][$eq]=" + splitVal[i].trim();
+      // Remove active learners from generators (on ac page we have generators and classifiers but we want to exclude active learners there)
+      if (splitVal[i].trim() == 'generator' && this.executionTypeFilter != "activeLearner") {
+        filter += "&filters[executionType][$ne]=activeLearner";
+      }
```
SimonDegrafKern: Just an idea but in no means objective better: 

Line 145-148 could be outsourced in an own method with strong name in the signature so it could be straight up understandable what in the for loop happens.

Something e.g.

for [...] {
  filter += createFilterCriteria(attribute)
}

src/app/base/components/bricks-integrator/bricks-integrator.component.ts
```diff
@@ -172,7 +176,10 @@ export class BricksIntegratorComponent implements OnInit, OnDestroy {
         if (this.config.extendedIntegrator) {
           finalData = data.data.map(e => {
             const c = jsonCopy(e);
```
SimonDegrafKern: What does "e" and "c" mean in this context? I could maybe search it up in the broader context but would prefer to have meaningfull names here :)

src/app/base/components/bricks-integrator/bricks-integrator.component.ts
```diff
@@ -363,6 +368,7 @@ export class BricksIntegratorComponent implements OnInit, OnDestroy {
         else {
           this.config.api.data = c;
           this.config.api.data.data.attributes.partOfGroup = JSON.parse(c.data.attributes.partOfGroup);
+          this.config.api.data.data.attributes.partOfGroup = filterArrayMinusExcluded(this.config.api.data.data.attributes.partOfGroup, REMOVE_GROUP_NODES);
           this.config.api.data.data.attributes.partOfGroupText = this.config.api.data.data.attributes.partOfGroup.map(x => this.getGroupName(x)).join(", ");
           this.config.api.data.data.attributes.availableFor = JSON.parse(c.data.attributes.availableFor);
           this.config.api.data.data.attributes.integratorInputs = JSON.parse(c.data.attributes.integratorInputs);
```
SimonDegrafKern, line 380: This could maybe outsourced in a method which has a name that shows what this else case is about. So I see that this else case happens if integratorInputs are not set (at least thats what I undestand) but I don't know straight up what then happens in the else case. Is it more of an error handling or setting of default values or something in this direction? I think the implicit knowledge which is hidden here in the else case could be expressed explicitly for the reader in a method name. But this is again maybe more of code style and could be applied to mulitple places in our app, so again, you have to decide :)

submodules/javascript-functions/general.ts
lumburovskalina: I would that **filterArray** may be too general because the filter can be done on different conditions.
What do think about this:
`function removeArrayFromArray(mainArray: any[], arrayToRemove: any[])`
or
`function excludeArrayFromArray(mainArray: any[], arrayToExclude: any[])`

src/app/base/components/bricks-integrator/bricks-integrator.component.ts
lumburovskalina: i would personally not use a method here because what we do here is field parsing and removing the gdpr fields and after that we use the same field  `this.config.api.data.data.attributes` to set some additional data. If I take a look at this at once I would understand what is does and I have all the preparations at one place. If I had to look in another method and come back to see the other fields i would probably not remember all fields at once and will have to take a look at it again. 
Of course if these are used on multiple places i would put them in a method, but for now i would say we keep it like this :)

submodules/javascript-functions/general.ts
SimonDegrafKern: I agree that my suggestion is pretty generic and maybe not optimal. Out of the two you proposed, I think the first is clearer to me since remove is common for me in contrast to exclude as a programming term. But still, this is kinda subjective, so, please decide on the option you like the most. It was important for me to give you a perspective on how I understood the signature of the method :)

src/app/base/components/bricks-integrator/bricks-integrator.component.ts
SimonDegrafKern: I understand, then maybe add a comment at the top of the else-case body, with a short description of what is happening there ?

submodules/javascript-functions/general.ts
lumburovskalina: Okay, let's name it as the first suggestion

src/app/base/components/bricks-integrator/bricks-integrator.component.ts
lumburovskalina: Yes sure I will add a comment

submodules/javascript-functions/general.ts
SimonDegrafKern: fits fine :)

JWittmeyer:
I think there is something wrong here 

image

- [x] resolved

JWittmeyer:
I know we talked about this already but I think there is still some misconception.
image

This error is shown in the record ide when I select an active learner. They shouldn't be selectable in the record ide.

- [x] resolve

Also since active learner in general currently don't work with the bricks integrator I would have expected a note in the PR description and (if possible) link to the bricks pr that fixes it (discussion with Leo)

Lina: okay, i understood that we shouldn't display them on the attribute calculation page but on record ide and AL they should stay, then the condition you will se in the bricks integrator is wrong.

src/app/base/components/bricks-integrator/helper/dummy-nodes.ts
```diff
@@ -215,4 +215,8 @@ def vader_sentiment(record):
         elif vs["compound"] <= -0.05:
             return "negative"
     elif MODE == "scores": 
-        return vs`;
\ No newline at end of file
+        return vs`;
+
+
+// TODO: Remove this when grdp is removed from bricks
+export const REMOVE_GROUP_NODES = ['gdpr_compliant', 'not_gdpr_compliant'];
```
JWittmeyer: personal prefernce:
Since this is a const I'd go with GROUPS_TO_REMOVE or BLACKLISTED_GROUPS since REMOVE_GROUP_NODES sounds like a function. Yes, the capitalization should also "hint" at a const value and it's also temporary so no real problem.

SimonDegrafKern (approved):

PR #149, opened 2023-07-13
Custom project name for project snapshot and sample projects 

### LINKED ISSUE #261: Project snapshot name can be set when uploading a new snapshot
**Is your feature request related to a problem? Please describe.**
If we want to create a new project with uploading new records, we have a validation that we cannot use the same project name twice, but we export the project and import it again then we have projects with the same name. I noticed that in gates it’s hard to find your project since imported projects and sample projects can have the same name.

**Describe the solution you'd like**
We can have an optional input field in the upload component for snapshots (and maybe lookup lists) where the user can enter the desired name (same validation as we have when creating a new project with records).

**Describe alternatives you've considered**
 We could also think for a differentiation of the sample projects names (index of the sample project, or entering a name before submitting the project..).

lumburovskalina:
Closes: https://github.com/code-kern-ai/refinery/issues/261
PRs:
- https://github.com/code-kern-ai/refinery-gateway/pull/140
- https://github.com/code-kern-ai/refinery-ui/pull/149
- https://github.com/code-kern-ai/refinery-submodule-model/pull/47

### DIFF
--- a/src/app/base/services/project/project-mutations.ts
+++ b/src/app/base/services/project/project-mutations.ts
@@ -12,7 +12,7 @@ export const mutations = {
   `,
   CREATE_SAMPLE_PROJECT: gql`
   mutation($name:String, $projectType: String){
-    createSampleProject(name:$name, projectType:$projectType){
+    createSampleProject(name: $name, projectType: $projectType){
       ok
       project{
         id
--- a/src/app/base/services/project/project-mutations.ts
+++ b/src/app/base/services/project/project-mutations.ts
@@ -11,8 +11,8 @@ export const mutations = {
     }
   `,
   CREATE_SAMPLE_PROJECT: gql`
-  mutation($name:String){
-    createSampleProject(name:$name){
+  mutation($name:String, $projectType: String){
+    createSampleProject(name:$name, projectType:$projectType){
       ok
       project{
         id
--- a/src/app/base/services/project/project-mutations.ts
+++ b/src/app/base/services/project/project-mutations.ts
@@ -2,7 +2,7 @@ import { gql } from 'apollo-angular';
 
 export const mutations = {
   CREATE_PROJECT: gql`
-    mutation ($name: String!, $description: String!) {
+    mutation ($name: String, $description: String!) {
       createProject(name: $name, description: $description) {
         project {
           id
--- a/src/app/import/components/modal-upload/modal-upload.component.html
+++ b/src/app/import/components/modal-upload/modal-upload.component.html
@@ -9,7 +9,7 @@
         {{subTitle}}
     </div>
     <div class="form-control" *ngIf="uploadFileType == UploadFileType.PROJECT">
-        <label class="text-gray-500 text-sm font-normal">Project title<em>- optional</em></label>
+        <label class="text-gray-500 text-sm font-normal">Project title<em> - optional</em></label>
         <input [value]="projectName" (input)="setProjectName($event.target.value)" type="text"
             placeholder="Enter some title here ..." required autocomplete="off"
             class="input input-sm input-bordered w-full placeholder-italic">
@@ -24,11 +24,8 @@
                 <path d="M12 9v4" />
                 <path d="M12 17h.01" />
             </svg>
-            <label class="text-yellow-700 text-xs italic ml-2">If a project title is not specified, the project title
-                from
-                the
-                file
-                will be used, potentially leading to duplicate project names.</label>
+            <label class="text-yellow-700 text-xs italic ml-2">If no project title is specified, the title mentioned in
+                the uploaded data will be used, potentially leading to a duplication of project titles.</label>
         </div>
     </div>
 
--- a/src/app/import/components/modal-upload/modal-upload.component.html
+++ b/src/app/import/components/modal-upload/modal-upload.component.html
@@ -8,12 +8,20 @@
     <div class="text-sm text-gray-500 mb-4 text-center">
         {{subTitle}}
     </div>
+    <div *ngIf="file" class="form-control">
+        <label class="text-gray-500 text-sm font-normal">Project title</label>
+        <input [value]="projectName" (input)="setProjectName($event.target.value)" type="text"
+            placeholder="Enter some title here ..." required autocomplete="off"
+            class="input input-sm input-bordered w-full placeholder-italic">
+    </div>
+
     <kern-upload [uploadFileType]="uploadFileType" [uploadOptions]="{
             deleteProjectOnFail: uploadOptions.deleteProjectOnFail,
             reloadOnFinish: uploadOptions.reloadOnFinish,
             knowledgeBaseId: uploadFileType == UploadFileType.KNOWLEDGE_BASE ? uploadOptions.knowledgeBaseId : null,
             isModal: uploadOptions.isModal,
             tokenizer: uploadOptions.tokenizer,
-            showBadPasswordMsg: uploadOptions.showBadPasswordMsg
+            showBadPasswordMsg: uploadOptions.showBadPasswordMsg,
+            projectName: projectName
         }" (fileAttached)="setFile($event)" (badPasswordMsg)="setBadPasswordMsg($event)"></kern-upload>
 </kern-modal>
--- a/src/app/import/components/modal-upload/modal-upload.component.html
+++ b/src/app/import/components/modal-upload/modal-upload.component.html
@@ -8,11 +8,28 @@
     <div class="text-sm text-gray-500 mb-4 text-center">
         {{subTitle}}
     </div>
-    <div *ngIf="file" class="form-control">
-        <label class="text-gray-500 text-sm font-normal">Project title</label>
+    <div class="form-control" *ngIf="uploadFileType == UploadFileType.PROJECT">
+        <label class="text-gray-500 text-sm font-normal">Project title<em>- optional</em></label>
         <input [value]="projectName" (input)="setProjectName($event.target.value)" type="text"
             placeholder="Enter some title here ..." required autocomplete="off"
             class="input input-sm input-bordered w-full placeholder-italic">
+        <div *ngIf="isProjectTitleDuplicate" class="text-red-700 text-xs mt-2">Project title exists</div>
+        <div class="flex flex-row mt-2">
+            <svg xmlns="http://www.w3.org/2000/svg" class="icon icon-tabler icon-tabler-alert-triangle text-yellow-700"
+                width="20" height="20" viewBox="0 0 24 24" stroke-width="1.5" stroke="currentColor" fill="none"
+                stroke-linecap="round" stroke-linejoin="round">
+                <path stroke="none" d="M0 0h24v24H0z" fill="none" />
+                <path
+                    d="M10.24 3.957l-8.422 14.06a1.989 1.989 0 0 0 1.7 2.983h16.845a1.989 1.989 0 0 0 1.7 -2.983l-8.423 -14.06a1.989 1.989 0 0 0 -3.4 0z" />
+                <path d="M12 9v4" />
+                <path d="M12 17h.01" />
+            </svg>
+            <label class="text-yellow-700 text-xs italic ml-2">If a project title is not specified, the project title
+                from
+                the
+                file
+                will be used, potentially leading to duplicate project names.</label>
+        </div>
     </div>
 
     <kern-upload [uploadFileType]="uploadFileType" [uploadOptions]="{
--- a/src/app/projects/components/projects/projects-helper.ts
+++ b/src/app/projects/components/projects/projects-helper.ts
@@ -1,5 +1,6 @@
 export type ProjectsModals = {
     uploadProject: UploadProjectModal;
+    projectNameSampleProject: ProjectNameSampleProjectModal;
 };
 
 export type UploadProjectModal = {
@@ -7,11 +8,20 @@ export type UploadProjectModal = {
     doingSomething: boolean;
 };
 
+export type ProjectNameSampleProjectModal = {
+    open: boolean;
+    projectNameExists: boolean;
+}
+
 export function createDefaultProjectsModals(): ProjectsModals {
     return {
         uploadProject: {
             open: false,
             doingSomething: false,
+        },
+        projectNameSampleProject: {
+            open: false,
+            projectNameExists: false
         },
     };
 }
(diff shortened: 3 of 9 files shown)


src/app/base/services/project/project-mutations.ts
```diff
@@ -11,8 +11,8 @@ export const mutations = {
     }
   `,
   CREATE_SAMPLE_PROJECT: gql`
-  mutation($name:String){
-    createSampleProject(name:$name){
+  mutation($name:String, $projectType: String){
+    createSampleProject(name:$name, projectType:$projectType){
```
SimonDegrafKern:
```suggestion
    createSampleProject(name: $name, projectType: $projectType){
```

src/app/import/components/modal-upload/modal-upload.component.html
```diff
@@ -8,12 +8,37 @@ <h1 class="flex flex-grow justify-center text-lg text-gray-900 font-bold">{{titl
     <div class="text-sm text-gray-500 mb-4 text-center">
         {{subTitle}}
     </div>
+    <div class="form-control" *ngIf="uploadFileType == UploadFileType.PROJECT">
+        <label class="text-gray-500 text-sm font-normal">Project title<em>- optional</em></label>
```
SimonDegrafKern:
```suggestion
        <label class="text-gray-500 text-sm font-normal">Project title<em> - optional</em></label>
```

src/app/import/components/modal-upload/modal-upload.component.html
```diff
@@ -8,12 +8,37 @@ <h1 class="flex flex-grow justify-center text-lg text-gray-900 font-bold">{{titl
...
+                width="20" height="20" viewBox="0 0 24 24" stroke-width="1.5" stroke="currentColor" fill="none"
+                stroke-linecap="round" stroke-linejoin="round">
+                <path stroke="none" d="M0 0h24v24H0z" fill="none" />
+                <path
+                    d="M10.24 3.957l-8.422 14.06a1.989 1.989 0 0 0 1.7 2.983h16.845a1.989 1.989 0 0 0 1.7 -2.983l-8.423 -14.06a1.989 1.989 0 0 0 -3.4 0z" />
+                <path d="M12 9v4" />
+                <path d="M12 17h.01" />
+            </svg>
+            <label class="text-yellow-700 text-xs italic ml-2">If a project title is not specified, the project title
+                from
```
SimonDegrafKern, line 28: I am not 100% sure, but should it maybe be "of" instead of "from" ?

src/app/import/components/modal-upload/modal-upload.component.html
```diff
@@ -8,12 +8,37 @@ <h1 class="flex flex-grow justify-center text-lg text-gray-900 font-bold">{{titl
...
+                the
+                file
+                will be used, potentially leading to duplicate project names.</label>
```
SimonDegrafKern: I am not 100% sure if this is better, but still I would rephrase this to something like:

_If no project title is specified, the title mentioned in the uploaded data will be used, potentially leading to a duplication of project titles._

src/app/projects/components/projects/projects-helper.ts
```diff
@@ -1,17 +1,27 @@
 export type ProjectsModals = {
     uploadProject: UploadProjectModal;
+    projectNameSampleProject: ProjectNameSampleProjectModal;
```
SimonDegrafKern: The name makes sense, but is a bit long in my opinion. What do you think about the name **sampleProjectName**?

src/app/projects/components/projects/projects.component.html
```diff
@@ -354,7 +354,29 @@
...
 }" (closeModalEvent)="projectsModals.uploadProject.open = false;" (refetchProjects)="refetchProjects($event)"
     (badPasswordMsg)="setBadPasswordMsg($event)">
-</kern-modal-upload>
\ No newline at end of file
+</kern-modal-upload>
+
+<kern-modal [isOpen]="projectsModals.projectNameSampleProject.open" closeButton="X" [modalBoxStyle]="{'width':'35rem'}"
+    [acceptButton]="{
+    buttonCaption: 'Create',
+    closeAfterClick:false,
```
SimonDegrafKern:
```suggestion
    closeAfterClick: false,
```

src/app/projects/components/projects/projects.component.html
```diff
@@ -354,7 +354,29 @@
...
+
+        <div class="flex flex-row">
+            <input [value]="projectName" (input)="checkIfProjectNameExists($event.target.value)" type="text"
+                placeholder="Enter some title here ..." required autocomplete="off"
+                class="input input-sm input-bordered w-full placeholder-italic">
+        </div>
+        <div *ngIf="projectsModals.projectNameSampleProject.projectNameExists" class="text-red-700 text-xs mt-2">Project
+            title exists</div>
+    </div>
+</kern-modal>
```
SimonDegrafKern, line 398: In this modal the terms "title" and "name" are used. Is there any difference between them? Is there a reason why both are used? Can they maybe even be merged into one term?

src/app/projects/components/projects/projects.component.ts
```diff
@@ -168,10 +170,12 @@ export class ProjectsComponent implements OnInit, OnDestroy {
...
-          projectItem.timeStamp = parseUTC(projectItem.createdAt);
-          const splitDateTime = projectItem.timeStamp.split(',');
-          projectItem.date = splitDateTime[0].trim();
-          projectItem.time = splitDateTime[1];
+          const projectItemCopy = jsonCopy(projectItem);
+          projectItemCopy.timeStamp = parseUTC(projectItemCopy.createdAt);
+          const splitDateTime = projectItemCopy.timeStamp.split(',');
+          projectItemCopy.date = splitDateTime[0].trim();
+          projectItemCopy.time = splitDateTime[1];
+          projectItem = projectItemCopy;
```
SimonDegrafKern: Why is a copy used here? What would happen, if we just use the original projectItem object and change the date and time information directly on it?

src/app/projects/components/projects/projects.component.ts
```diff
@@ -272,4 +284,9 @@ export class ProjectsComponent implements OnInit, OnDestroy {
   setBadPasswordMsg(showBadPassMgs: boolean) {
     this.showBadPasswordMsg = showBadPassMgs;
   }
+
+  checkIfProjectNameExists(value: string) {
+    this.projectName = value;
+    this.projectsModals.projectNameSampleProject.projectNameExists = this.projectList.some(project => project.name == this.projectName);
+  }
```
SimonDegrafKern, line 290: I am not an expert in typeScript, so maybe this is just valid, but I am a bit confused why this method manipulates the instance variables on its own. This works just fine, but is a bit of a side effect to me, since the method name says checkIfProjectNameExists and not checkAndSetIfProjectNameExists. The difference for me is, that a checker-Method only does "read-only" operations and does not mutate anyhing while this method does. Like I said, it is not really a problem, but worthy to think of in terms of clean code. But like I already mentioned I am not really proficient in typeScript and could be that this is best practice in the language.

I would have expected a return of the result of the check.

SimonDegrafKern:
I was able to create two projects with the same name on sample project import by using a space after the title clickbait like "Clickbait ". Maybe there should be a trim() or a method which removes whitespaces be used in the check for project names? 

image


- [x] resolved

SimonDegrafKern:
When importing by using a project snapshot I could just duplicate a name even when there was a warning. Is this the expected behaviour?

@SimonDegrafKern with the current implementation we check if the file name exists as a name in the projects, and we cannot use the same name. But if the user does not provide a project title, then we use the name of the project in the file - that's the idea of the warning , in this case it can come to duplicate names. Ex. you export a 'Clickbait' project with Clickbait name and you try to import it without setting the name in the project title field. (that's what the warning tries to tell - if this is not clear we may need different message). 
Does this answer your question? Did you the following steps?

@lumburovskalina yeah thats what I meant. A way to avoid duplicated names could be to check this on the backend. In the backend we already know about the data in the file since we open it, so if a duplicated name gets recognized we could use the name of the file and append something like " (1)" .. and count up for the same name with e.g. " (2)" for it. What do you think about it? So a second clickbait project would then just be called "Clickbait (1)"

SimonDegrafKern:
image
image

When seeing both modals the one which requires a change of name looks a bit "empty" imo. I think this is because of the single big headline which has no description while the other modal has a description, which in my opinion looks/feels better. Don't know if this is relatable for you, maybe we can talk about this, but anyway no really important point but worth to mentio n

- [x] resolved
- [x] discussed

@SimonDegrafKern What do you think about this?
 
Screenshot 2023-07-19 at 11 13 29

@lumburovskalina this makes it look better for me :)

src/app/import/components/modal-upload/modal-upload.component.html
lumburovskalina: Yes, this looks good.

src/app/projects/components/projects/projects-helper.ts
lumburovskalina: Yes, I agree (two times project in the name)

src/app/projects/components/projects/projects.component.html
lumburovskalina: No, there is no difference between the terms, we should use 'title'. I will replace 'Enter project name' to 'Enter project title' ( i think that was the only occurrence)

src/app/projects/components/projects/projects.component.ts
lumburovskalina: Some object in Typescript cannot be extended (in the query we specify which fields are coming from BE) and if we want to extend the object we need a copy and then add the new object to the list.

src/app/projects/components/projects/projects.component.ts
lumburovskalina: This function is called on (input) event so everytime the user is typing something we want to check if the name exists and set 'projectNameExists' to true. But, yes I agree it checks and sets so the name 'checkAndSetIfProjectNameExists' would make more sense.

src/app/projects/components/projects/projects.component.ts
SimonDegrafKern: Ah, that makes sense, thanks for the explanation

src/app/projects/components/projects/projects.component.ts
SimonDegrafKern: works for me :)

SimonDegrafKern (approved):

src/app/projects/components/projects/projects.component.ts
```diff
@@ -164,17 +166,16 @@ export class ProjectsComponent implements OnInit, OnDestroy {
...
-      this.projectList = projectList.filter(a => a.status != ProjectStatus.IN_DELETION);
+      this.projectList = projectList.map((project) => {
+        const projectItemCopy = jsonCopy(project);
+        projectItemCopy.timeStamp = parseUTC(projectItemCopy.createdAt);
+        const splitDateTime = projectItemCopy.timeStamp.split(',');
+        projectItemCopy.date = splitDateTime[0].trim();
+        projectItemCopy.time = splitDateTime[1];
+        return projectItemCopy;
+      })
+      this.projectList = this.projectList.filter(a => a.status != ProjectStatus.IN_DELETION);
```
JWittmeyer: we first map all and then filter? why not the other way around? 
```suggestion
      this.projectList = projectList.filter(a => a.status != ProjectStatus.IN_DELETION).map((project) => {
        const projectItemCopy = jsonCopy(project);
        projectItemCopy.timeStamp = parseUTC(projectItemCopy.createdAt);
        const splitDateTime = projectItemCopy.timeStamp.split(',');
        projectItemCopy.date = splitDateTime[0].trim();
        projectItemCopy.time = splitDateTime[1];
        return projectItemCopy;
      })
```

JWittmeyer: also i would sort the final list (this.projectList) not the input list

src/app/projects/components/projects/projects.component.ts
```diff
@@ -227,12 +228,20 @@ export class ProjectsComponent implements OnInit, OnDestroy {
     if (value == 'Further sample projects') {
       window.open("https://github.com/code-kern-ai/refinery-sample-projects", "_blank");
     } else {
-      this.importSampleProject(this.isProjectInitial ? this.previousValue + ' - initial' : value)
+      this.projectName = this.isProjectInitial ? this.previousValue + ' - initial' : value;
+      this.projectType = this.isProjectInitial ? this.previousValue + ' - initial' : value;
```
JWittmeyer: whats the difference if they are exactly the same?

src/app/projects/components/projects/projects.component.ts
```diff
@@ -272,4 +281,9 @@ export class ProjectsComponent implements OnInit, OnDestroy {
   setBadPasswordMsg(showBadPassMgs: boolean) {
     this.showBadPasswordMsg = showBadPassMgs;
   }
+
+  checkAndSetIfProjectNameExists(value: string) {
+    this.projectName = value;
+    this.projectsModals.sampleProjectName.projectNameExists = this.projectList.some(project => project.name == this.projectName.trim());
```
JWittmeyer: this would call the trim function on every entry. I'm assuming that we need to store the untrimmed value in projectName so i'd go with:
```suggestion
    this.projectName = value;
    const checkName = value.trim();
    this.projectsModals.sampleProjectName.projectNameExists = this.projectList.some(project => project.name == checkName);
```

JWittmeyer:
So far the logic for sample projects was -> create & if you are on the projects page navigate to the sample projects overview page
``` typescript
    this.projectApolloService.createSampleProject(this.projectName, this.projectType).pipe(first()).subscribe((p: Project) => {
      if (this.router.url == "/projects") { 
        this.router.navigate(['projects', p.id, 'overview']);
      }
    });
```

So far this was fine-ish since if we didn't have many interaction options on the projects page. Now with the extension I'd include the modal open states in the check so we don't navigate away and "lose" the project name. 
This should check the naming modal for sample projects as well as the snapshot modal. if they are open we don't navigate away

- [x] resolved

JWittmeyer:
remove button of snapshot import (after file was selected) doesn't clear the name field

- [x] resolved

JWittmeyer:
Multiple file endings (like from our project export) result in file ending in names. 
image

I think i'd go with a loop to remove the "dot" parts

- [x] resolved

src/app/projects/components/projects/projects.component.ts
lumburovskalina: yes, filter before map should be the correct behaviour

src/app/projects/components/projects/projects.component.ts
lumburovskalina: This is for the first creation of the project (the name and the type are the same), but I can just assign the value, no need to calculate the values again.

SimonDegrafKern (approved):

PR #150, opened 2023-07-20
Qdrant filter integrations

lumburovskalina:
PRs:
- https://github.com/code-kern-ai/refinery-ui/pull/150
- https://github.com/code-kern-ai/refinery-gateway/pull/142
- https://github.com/code-kern-ai/refinery-submodule-model/pull/49
- https://github.com/code-kern-ai/refinery-neural-search/pull/31
- https://github.com/code-kern-ai/refinery-embedder/pull/52

### DIFF
--- a/src/app/base/components/dropdown/dropdown.component.html
+++ b/src/app/base/components/dropdown/dropdown.component.html
@@ -12,7 +12,8 @@
             }" [ngClass]="buttonClassList" (isMenuOpen)="toggleVisible($event, dropdownOptionsDiv)" appDropdown>
             {{ dropdownOptions.hasCheckboxes?'':dropdownOptions.buttonCaption }}
             <ng-template [ngIf]="dropdownOptions.hasCheckboxes">
-                <label class="truncate cursor-pointer">{{getDropdownDisplayText(dropdownOptions.optionArray,"EMPTY")}}
+                <label
+                    class="truncate cursor-pointer text-sm">{{getDropdownDisplayText(dropdownOptions.optionArray,"EMPTY")}}
                     <span
                         style="color:#2563eb">{{getDropdownDisplayText(dropdownOptions.optionArray,"NOT_NEGATED")}}</span>
                     <span style="color:#ef4444">{{getDropdownDisplayText(dropdownOptions.optionArray,"NEGATED")}}</span>
--- a/src/app/base/components/dropdown/dropdown.component.html
+++ b/src/app/base/components/dropdown/dropdown.component.html
@@ -40,13 +40,14 @@
     </a>
 
     <div #dropdownOptionsDiv
-        class="origin-top-right absolute mt-2 rounded-md shadow-lg bg-white ring-1 ring-black ring-opacity-5 focus:outline-none hidden"
+        class="origin-top-right mt-2 rounded-md shadow-lg bg-white ring-1 ring-black ring-opacity-5 focus:outline-none hidden"
         [ngClass]="[dropdownClassList, dropdownOptions.isButtonSampleProjects ? 'width-sample-projects' : '', dropdownOptions.buttonVersion != '...' ? 'min-w-full': '', dropdownOptions.addScrollXDropdownOptions ? 'overflow-x-auto' : '', dropdownOptions.maxWidth ? 'overflow-y-auto' : '']"
         [style.max-height]="dropdownOptions.maxHeight ? dropdownOptions.maxHeight : ''"
         [style.overflow-y]="dropdownOptions.maxHeight ? 'auto' : null"
         [style.max-width]="dropdownOptions.maxWidth ? dropdownOptions.maxWidth : ''" role="menu"
         [style.width]="dropdownOptions.width ? dropdownOptions.width : ''" aria-orientation="vertical"
-        aria-labelledby="menu-button" tabindex="-1">
+        [style.position]="dropdownOptions.useRelativePosOptions ? 'relative' : 'absolute'" aria-labelledby="menu-button"
+        tabindex="-1">
         <div class="py-1 cursor-pointer" role="none">
             <div *ngFor="let caption of dropdownOptionCaptions;let i = index" (click)="performActionOnOption($event,i)"
                 [attr.data-tip]="dropdownOptions.optionTooltips && dropdownOptions.optionTooltips[i]?dropdownOptions.optionTooltips[i]:null"
--- a/src/app/base/components/dropdown/dropdown.component.scss
+++ b/src/app/base/components/dropdown/dropdown.component.scss
@@ -6,6 +6,10 @@
     width: 384px !important;
 }
 
+.tooltip_max_width_150::before {
+    max-width: 150px;
+}
+
 .tooltip_max_width_200::before {
     max-width: 200px;
 }
--- a/src/app/base/components/dropdown/dropdown.component.ts
+++ b/src/app/base/components/dropdown/dropdown.component.ts
@@ -61,7 +61,7 @@ export class DropdownComponent implements OnChanges {
 
   private buildHelperValues() {
     this.buttonClassList += this.dropdownOptions.isDisabled ? 'opacity-50 cursor-not-allowed ' : 'opacity-100 cursor-pointer ';
-    this.dropdownClassList = this.dropdownOptions.hasCheckboxes ? ' w-80 ' : '';
+    this.dropdownClassList = this.dropdownOptions.hasCheckboxes && !this.dropdownOptions.hasFullWidth ? ' w-80 ' : '';
     this.dropdownClassList += this.dropdownOptions.buttonVersion != 'default' ? 'right-0 width-icon-menues' : '';
     this.buttonClassList += this.dropdownOptions.isButtonSampleProjects ? 'py-2' : 'border-gray-300 py-1.5';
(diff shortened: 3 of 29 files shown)


src/app/base/components/dropdown/dropdown.component.html
```diff
@@ -40,13 +40,14 @@
     </a>
 
     <div #dropdownOptionsDiv
-        class="origin-top-right absolute mt-2 rounded-md shadow-lg bg-white ring-1 ring-black ring-opacity-5 focus:outline-none hidden"
+        class="origin-top-right mt-2 rounded-md shadow-lg bg-white ring-1 ring-black ring-opacity-5 focus:outline-none hidden"
```
SimonDegrafKern, line 44: I don't know that much about css classes, but the last class used here is hidden -> does this hide the whole element? Wouldn't the other classes then be unnecessary? Most probably the hidden does something different, but since I am a total noob in css I am interested in what this does :D

src/app/base/components/dropdown/dropdown.component.scss
```diff
@@ -6,6 +6,10 @@
     width: 384px !important;
 }
 
+.tooltip_max_width_150::before {
```
SimonDegrafKern, line 9: why is tooltip here in the name? Could this class not be used for every element and not only tooltips?

src/app/base/components/dropdown/dropdown.component.ts
```diff
@@ -218,6 +218,9 @@ export class DropdownComponent implements OnChanges {
         c.get('active').value &&
         (!hasNegate || (hasNegate && !c.get('negate').value))
       ) {
+        if (this.dropdownOptions.hasSelectAllOption) {
+          if (c.get('id').value == 'SELECT_ALL') continue;
+        }
```
SimonDegrafKern, line 223: The logic in this for loop seems to be quite complex for me with many if-elses and nested cases. I would think of maybe refactoring to use the same structure but outsource the checks into expressive functions so it is easy understandable what happens in this loop :) This is like always my opinion, but to give you an idea what my perspective is:

I am not that familiar with our frontend development at all but what I see is something like this (Attention: Javascript like peusdocode :D) :

```
if (labelFor == 'EMPTY' && c.get('active').value) ...
else if (
  labelFor == 'NOT_NEGATED' &&
  c.get('active').value &&
  (!hasNegate || (hasNegate && !c.get('negate').value))
) ... 
else if (
  labelFor == 'NEGATED' &&
  c.get('active').value &&
  hasNegate &&
  c.get('negate').value
) ... 
  if (
    !atLeastOneNegated &&
    c.get('active').value &&
    hasNegate &&
    c.get('negate').value
  ) ....
```

this could be rewritten into something like:
```
if(isActiveControlEmpty(x,y)) ...
elseif(isActiveControlNotNegate(x,y)) ...
elseif(isActiveControlNegated(x,y)) ...
else
```

I hope this is understandable :D So for me it would be way easier to understand what this method and the steps should do if it is written in a more (simon-) readable way, but like I said, just a perspective which can or can not be relevant for you :)

src/app/base/components/dropdown/dropdown.component.html
lumburovskalina: I removed the 'absolute' class from the dropdown because I added it as an optional parameter to the dropdown options:
`[style.position]="dropdownOptions.useRelativePosOptions ? 'relative' : 'absolute'"`
So, again it's default as absolute but if the user needs it for some specific dropdown as relative, he can use it by sending the field of the options as true (I needed it for one dropdown :D )

src/app/base/components/dropdown/dropdown.component.scss
lumburovskalina: This was a little bit tricky to add a max width to the tooltip, so we decided to go with class names that have different max width and those field is also part of the dropdown options. The name of the class is passed as string and that variable is part of the style of the element.

src/app/data/components/data-browser/helper-classes/filter-integration.ts
```diff
@@ -0,0 +1,161 @@
...
+    operatorsDict: { [key: string]: string[] } = {};
+    tooltipsDict: { [key: string]: string[] } = {};
+
+    constructor(dataBrowser: DataBrowserComponent, private formBuilder: FormBuilder) {
+        this.dataBrowser = dataBrowser;
+        this.formBuilder = formBuilder;
+    }
+
+    initFilterForm() {
+        this.filterAttributesSSForm = this.formBuilder.group({
```
SimonDegrafKern: What does SS stand for ? :D

lumburovskalina: Similarity Search :) 
in the beginning I thought we would have two forms: one for the similarity search, one for the outliers, but this is not the case anymore, I could remove it from the name so it doesn't make confusion

SimonDegrafKern: For me both would be okay changing it or keeping it, what is your opinion, is it clear from the context that similarity search is meant?

src/app/base/components/dropdown/dropdown.component.scss
SimonDegrafKern: Ah okay that makes sense. This can of course stay as it is but still I am confused why it says ".tooltip_max_width_150::before"  and not ".max_width_150::before", so it could be reused for other components. Or is it that the name itself gets compared in some way?

src/app/base/components/dropdown/dropdown.component.html
SimonDegrafKern: Ah okay thanks for explaining :)

src/app/base/components/dropdown/dropdown.component.scss
lumburovskalina: These are hard coded values that we use only for the tooltips (the before selector is the tooltip that pops up). We could of course add them in the global style and 'reuse' the class on other places, that wouldn't be wrong, but there is no need for that.
If it was a class 'max_width_150' and we have the max width: 150px on multiple places it would make sense to have it as part of the global styling (this could be solved by tailwind classes or in-line style -> we use combination of these two). Adding a class would make sense if we have multiple properties that we want to add to a specific element (also can be solved by multiple classes in Tailwind -> how we do it). But I wouldn't add a class with ::before in the global stying unless it is used everywhere. 
Also this approach with having the size (150) in the name of the class is only used here for the tooltips on the dropdown, so we can differentiate the classes. In other cases I would go with more descriptive class names.

src/app/data/components/data-browser/helper-classes/filter-integration.ts
lumburovskalina: Will change it :)

src/app/base/components/dropdown/dropdown.component.ts
lumburovskalina: I would keep it like this for now, maybe add some comments because I think we would do a lot of passing of the parameters for only checking the condition. Not sure if that's a better approach tho :)

src/app/base/components/dropdown/dropdown.component.scss
SimonDegrafKern: Alright, thanks for the explanation, I think I got it :)

src/app/base/components/dropdown/dropdown.component.ts
SimonDegrafKern: Okay, works for me :)

JWittmeyer:
Font style differs from other fields 

image

- [x] resolved

JWittmeyer:
Did the max with of the modal change?
(pr)
image

(dev)
image

- [ ] discussed
- [x] resolved

JWittmeyer:
Editing filter attributes:
- nothing to something
- something to nothing for azure

- [x] resolved

JWittmeyer:
Overflow is off for filter dropdown:
image


- [x] resolved

JWittmeyer:
Searching without filter should still be possible. The filters are completely optional

image

- [x] resolved

JWittmeyer:
The whole Embedding is recreated on payload change.
Better to just reupload the embedding values with new payload info for qdrant

- [x] resolved

JWittmeyer:
When i switch to the databrowser i get an error:

image

Seems to happen when no filter attributes are given for an embedding

- [x] resolved

JWittmeyer:
Searching with an embedding without filter attributes now results in issues:
image

- [x] resolved

JWittmeyer:
I think we don't need the second part anymore since we don't recreate the embedding. 


image

- [x] resolved

JWittmeyer:
Saving the filter attribute without changing them results in a reupload to qdrant. Maybe we can disable the save button if the values didn't change.

- [x] resolved

JWittmeyer:
Searching without filter still sends a filter wich results in faulty values.
In this case it tries to find similar records with attribute_4 as "" which is always nothing

image

I think the biggest challenge here is giving the option to search without filter & search with e.g. attribute= "" both are valid depending on the context. 

Maybe we use two buttons? Search & Search without a filter. I think for this we would need to redo the modal buttons since two accept options aren't planned but should be too complicated. 


- [x] resolved

JWittmeyer (approved):

PR #155, opened 2023-09-04
Option to edit records

JWittmeyer:
- https://github.com/code-kern-ai/refinery-ui/pull/155
- https://github.com/code-kern-ai/refinery-gateway/pull/150
- https://github.com/code-kern-ai/refinery-embedder/pull/60

best with `bash start -b edit-option`

caution for Mac Users, refinery-embedder is part of this PR which can result in some performance issues with the branch build
maybe then `bash start -b edit-option -e refinery-embedder` + embedder by hand

### DIFF
--- a/src/app/app-routing.module.ts
+++ b/src/app/app-routing.module.ts
@@ -62,8 +62,8 @@ const routes: Routes = [
       { path: 'labeling', component: LabelingSuiteComponent, data: { name: 'LabelingSuiteComponent' } },
       { path: 'labeling/:id', component: LabelingSuiteComponent, data: { name: 'LabelingSuiteComponent' } },
       // works with session id from data browser
-      { path: 'edit-records/:id', component: EditRecordsComponent, data: { name: 'EditRecordsComponent' }, canDeactivate: [CanDeactivateGuard] },
       { path: 'record-ide/:id', component: RecordIDEComponent, data: { name: 'RecordIDEComponent' } },
+      { path: 'edit-records', component: EditRecordsComponent, data: { name: 'EditRecordsComponent' }, canDeactivate: [CanDeactivateGuard] },
       { path: 'lookup-lists', component: KnowledgeBasesComponent, data: { name: 'KnowledgeBasesComponent' } },
       {
         path: 'lookup-lists/:knowledgeBaseId',
--- a/src/app/app-routing.module.ts
+++ b/src/app/app-routing.module.ts
@@ -21,6 +21,8 @@ import { ModelCallbackComponent } from './model-callbacks/components/model-callb
 import { UsersComponent } from './users/components/users.component';
 import { CrowdLabelerDetailsComponent } from './weak-supervision/components/crowd-labeler-details/component/crowd-labeler-details.component';
 import { ProjectAdminComponent } from './project-admin/component/project-admin.component';
+import { EditRecordsComponent } from './edit-records/components/edit-records.component';
+import { CanDeactivateGuard } from './util/CanComponentDeactivate';
 
 const routes: Routes = [
   {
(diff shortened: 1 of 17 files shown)


src/app/app-routing.module.ts
```diff
@@ -59,6 +61,8 @@ const routes: Routes = [
       { path: 'admin', component: ProjectAdminComponent, data: { name: 'ProjectAdminComponent' } },
       { path: 'labeling', component: LabelingSuiteComponent, data: { name: 'LabelingSuiteComponent' } },
       { path: 'labeling/:id', component: LabelingSuiteComponent, data: { name: 'LabelingSuiteComponent' } },
+      // works with session id from data browser
+      { path: 'edit-records/:id', component: EditRecordsComponent, data: { name: 'EditRecordsComponent' }, canDeactivate: [CanDeactivateGuard] },
```
FelixKirschKern: is the id used anywhere?

src/app/edit-records/components/edit-records.component.html
```diff
@@ -0,0 +1,383 @@
...
+            <div class="flex flex-row flex-nowrap items-center">
+                <div class="flex justify-center overflow-visible items-center">
+                    <div class="tooltip tooltip-left" data-tip="Persist changes">
+                        <button tabindex="0" (click)="openSyncModal();"
+                            class="bg-white text-gray-700 text-xs font-semibold mr-3 px-4 py-1.5 rounded-md border gap-x-1 border-gray-300 whitespace-nowrap inline-flex items-center hover:bg-gray-50 focus:outline-none">
+                            Synchronize with DB
+                        </button>
+                    </div>
+                    <div class="text-sm leading-5 text-gray-500 flex-shrink-0 mr-3 my-3">
+                        <div data-tip="If want different records, you can select new filters in the browser"
```
FelixKirschKern:
```suggestion
                        <div data-tip="If you want different records, you can select new filters in the browser"
```

src/app/edit-records/components/edit-records.component.ts
```diff
@@ -0,0 +1,211 @@
...
+      case "grid-cols-1": this.erd.columnClass = "grid-cols-2"; break;
+      case "grid-cols-2": this.erd.columnClass = "grid-cols-3"; break;
+      case "grid-cols-3": this.erd.columnClass = "grid-cols-1"; break;
+      default: this.erd.columnClass = "grid-cols-3";
+    }
+
+    localStorage.setItem("ERcolumnClass", this.erd.columnClass);
+    this.erd.data.selectedRecordId = null;
+    if (this.erd.editRecordId) timer(100).subscribe(() => this.erd.data.selectedRecordId = this.erd.editRecordId);
+    scrollElementIntoView("flash-it", 50);
```
FelixKirschKern: What is the difference between selectedRecordId and editRecordId?

When I change between column classes the record I am currently working on is not scrolled into view.
As far as I understood, the scrollElementIntoView scrolls to the element with id "flash-it", which is the element with the record id that equals the selectedRecordId. But selectedRecordId is null at the moment we're calling scrollElementIntoView, and because the timeout of scrollIntoView (50) is smaller than the one of the setter of the selectedRecordId (100), it's still null when the scrollIntoView is called.

I also get the console warning: "scrollElementIntoView elementId not found".

src/app/app-routing.module.ts
JWittmeyer: removed

src/app/edit-records/components/edit-records.component.html
JWittmeyer: changed

src/app/edit-records/components/edit-records.component.ts
JWittmeyer: long story short selected is the record id that is either set on opening the page via the databroser or after changing the layout to the currently "open" record. The flash-it animation is id-based and for the animation to take effect it needs a switch from no id to id that's why there is a timer involved.

-> changed to if else logic

FelixKirschKern (approved):

